PR #1, opened 2019-12-04
Add Rich Image Enhancements

mkaz:
The rich image filter adds additional operations such as flip, rotate, and filters to an image.

@johngodley - The rest-api.php part was a little tricky. I tweaked it and I think its working now (for this one block) but I kinda hacked in. Also the top-level rest-api (copied into the block) still looks like it is based of Pages namespace.

Feel free to tweak on this PR and merge when you think its ready.


johngodley:
I've split the API code out a bit more, and merged it in.

PR #6, opened 2019-12-12
Fix the grid block resize not working in the 7.1 editor

johngodley:
A change in the inner blocks class name broke the CSS.

`editor-block-list__layout` => `block-editor-block-list__layout`

This supports both, so it should work on any version of Gutenberg.

I note that elsewhere we use `block-editor-block-list__layout`. Do we even need to support `editor-block-list__layout`?


jasmussen:
> editor-block-list__layout

This is the old label, and apparently we forgot to remove it, I wonder if there were deprecation warnings. the "editor" prefix is gone, and everything is "block-editor" now. My bad, apologies.

If this fixes it feel free to merge, otherwise I can take a look in 5 minutes

PR #7, opened 2019-12-12
Fix layout grid in 7.1 and specifically nested in groups.

### LINKED ISSUE #4: Layout Grid: Content overflows when it's inside a Group block
(copied from elsewhere)

A group block is handy to add a background colour or give a sense of "section".  When I add a Layout Grid block into a Group block, the overflow issue comes back. And a horizontal scroll starts to appear in the editor.

wordpress com_block-editor_post_iamtakashinuxtest2019112113719 wordpress com_301 (1)

Yes, I'm aware of that Layout Grid block allows me to add background colours to its columns, but at the moment, it's not possible to have a background colour that spans edge-to-edge. (See no. 4 as an example)

I love the Layout Grid block as it works a lot better than Core's grid block. The block might have not been designed for being as an inner block, but it'd be nice if we can fix this.

jasmussen:
Fixes #4.

This fix remains a bit hacky, and I'll make a note to self to ponder better solutions. But if this works well in older Gutenberg versions, it would be good to ship as an interim fix.

### DIFF
--- a/blocks/layout-grid/editor.scss
+++ b/blocks/layout-grid/editor.scss
@@ -78,16 +78,6 @@
 			width: 100%;
 		}
 	}
-
-	&.is-selected > .block-editor-block-list__block-edit {
-		// When the block is selected, we scale it down to make room for the side UI.
-		width: calc( 100% + 44px + 44px );
-
-		// Make an exception for the group, which zeroes out a margin we need.
-		.wp-block-group & {
-			width: calc( 100% - 14px - 14px );
-		}
-	}
 }
 
 // Try hiding the alignments dropdown.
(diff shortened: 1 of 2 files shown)


johngodley:
Seems to work great. I've included screenshots from 6.7.0, and one from 7.1.0 - there is a difference in the unselected appearance. If that's ok then let's go with this.

**6.7.0**

Group + grid followed by grid. Both unselected

Edit_Page___Pages_by_WordPress_com

When grid in group is selected:

Edit_Page___Pages_by_WordPress_com_and_1Password

When grid is selected:

Edit_Page___Pages_by_WordPress_com_and_1Password

**7.1.0**

Group + grid followed by grid. Both unselected

Edit_Page___Pages_by_WordPress_com

jasmussen:
Thank you John, you helped surface some faulty CSS. The recent PR fixes it for the group block as well. I think we should ship this.

PR #16, opened 2019-12-28
Add Strike That block

mkaz:
A block that scribbles out the content of itself, using an SVG animation, [inspired by this page](https://wattenberger.com/blog/react-hooks)

strike-anim


jasmussen:
Oh this is cool! 

Can you elaborate a little bit on exactly what it does? Do you paint the scribble yourself? What triggers it? What are our options here?

mkaz:
@jasmussen I did paint the scribble myself, by scribbling with a pen and exporting as an SVG. It did a weird thing with the outline, which I like but we can probably improve on. The technique is animate is pure CSS, using I believe a trick you talked about but I [borrowed implementation from here](https://css-tricks.com/svg-line-animation-works/)

As far as options, we probably have quite a few. though not sure how useful it is overall.

jasmussen:
See when I look at the code I see this:

```
.wp-block-jetpack-strike-that svg {
	position: absolute;
	fill: none;

	path {
		stroke-dasharray: 12400;
		stroke-dashoffset: 12400;
		animation: dash 5s linear forwards;
	}

	@keyframes dash {
		to {
		stroke-dashoffset: 0;
		}
	}
}
```

This happens to be the same trick that powers the WordPress preview logo: https://codepen.io/joen/pen/jepZrB

And this old Mesh logo: https://codepen.io/joen/pen/RqmvKL

It's also what powers https://www.polygon.com/a/ps4-review

I think there may already be a ticket for it, but I feel like we could make a block that took ANY USER SVG (whether pasted in or uploaded) and applied this "draw the strokes" effect on it. Whether looping, on-scroll, or just "on appear". I think that would be a SWEET block. In a way, that could be this block, but with the SVG provided by the user, and less abs positioning and more of a square embed-style decorative treatment.

What do you think?

mkaz:
Closing. We can revisit if we like the idea.

PR #18, opened 2020-01-10
Add starscape block

ajlende:
## Description

Adds a starscape motion background block with a simple text entry.

Available options:
- Block alignment
- Text alignment
- Background gradient
- Star density
- Star rotation speed
- Max width & height (if needed, corners may be missing stars on 4k monitors with default settings, but doing the full 4k probably wouldn't perform as well on mobile)

starscape-walkthrough

## Screenshots

wpcomblocks local__p=757 preview=true (2)
wpcomblocks local__p=757 preview=true (1)
wpcomblocks local__p=757 preview=true


pablohoneyhoney:
Really good stuff!

Here something that can help

**1. Functionally**
Ideally we can limit the options, and offer good color defaults. Close to this:

Screen Shot 2020-01-20 at 2 28 44 PM


**2. Verbally**

**Editor**
Title: _Starscape Block_
Block directory title: _Starscape_
Subcopy in editor: _Create content with stars in motion._

To reduce copy in the sidebar to the basic:

Current: _The maximum dimensions control how far out to render stars. It does not change the size of the container. Keep these values as small as possible, especially with higher density stars, for better performance on low-powered devices._

New: _Control the area of stars you want. The smaller the values, the better the performance._ 

**Plugin**
Title: _Starscape Block_
Tagline: _Stars in motion_
Description: _Everything was made of collapsing stars, we are all made of star stuff. Now we also can create content in WordPress with stars in motion._



**3. Visually** 
Added a bit of personality to the block for promo purposes.
Ideally the block starts similar to the mockups once inserted.
Assets here: https://www.figma.com/file/2NV16ik4XcdOToBsydFqvL/Starscape-Block?node-id=3189%3A8541

Logo

Logo-1

Hero

Editor

Assets

ajlende:
@pablohoneyhoney I love the icon and graphics you added!

Just a couple questions/comments:

**1. Functionality**

a. I currently have the theme colors, plus black and white as available text colors. Should I make that palette a fixed set of colors to match the colors in the mockups instead?

b. Custom gradients aren't supported yet, unfortunately, so I want to still give enough flexibility. If I cut the number of choices in half—to just 12—would you say that's still too many?

**2. Verbally**

I think I need to clarify the purpose of the dimensions option better.

Simply saying a smaller number is better seems to imply that the obvious thing to do is to just make them zero for best performance. While that would perform the best, you also wouldn't see any stars.

More realistically, let's say you're editing on mobile or on some other smaller-than-average screen. You lower the numbers because you want better performance. It looks good on your screen, but upon opening it up on desktop, you see that there aren't any stars at the top, and there is an obvious arc where they stop which doesn't look good.

On the flip-side if you optimize for an 8k screen to future-proof the whole thing, when loading the page on mobile, most of those stars won't ever appear on screen, but the phone would still have to download and parse an extra few kB of CSS, and the animation may be jittery.

The default values I have set work fine up to a full-height starscape on a 1080p Full-HD screen. In my testing, bumping it up to 4k caused jank on mobile. If people really care about 4k and don't care at all about mobile, they can increase the values. If the site is running almost exclusively on mobile those number can be lowered so their mobile users use less data.

dimensions-explained

pablohoneyhoney:
I think 12 feels a more palatable range. We can always optimize from there. 

RE: dimensions, I understand, but as a good practice, we shouldn't rely on heavy copy to explain something like this in the sidebar. Can we limit the dimensions to optimized multi-screen ranges? Or can we edit the dimensions by device (desktop-mobile)?

ajlende:
Released on the WordPress.org plugin directory: https://wordpress.org/plugins/starscape/

PR #19, opened 2020-01-17
Add motion background block

ajlende:
## Checklist

- [x] 🎁 WebGL Block Template
- [x] 👾 Shared canvas for unlimited blocks (versus max 8-32 using multiple canvases)
- [x] 🖱 Mouse position interaction
- [x] 🖲 Scroll position interaction
- [x] ⏱ Time uniforms
- [x] 🖼 Image upload
- [x] 😎 Cool looking shades... er... shaders
- [x] 🔤 Text Heading
- [ ] 📦 Plugin/Block Directory Assets

## Name Ideas

- Liquid Image
- Paint Mixer

## Static Screenshots

Screen Shot 2020-02-12 at 18 11 32
Screen Shot 2020-02-12 at 18 11 49
Screen Shot 2020-02-12 at 18 11 42
Screen Shot 2020-02-12 at 18 11 38
\* Waves is just [a custom image](https://unsplash.com/photos/2mZavbuZfSY) and low complexity

## Add Block

paint-demo-2020-02-12-part1

## Gradient Colors

paint-demo-2020-02-12-part2

## Mouse Interaction & Animation Speeds

paint-demo-2020-02-12-part3

## Image Upload

paint-demo-2020-02-12-part4

## Curling Complexity

paint-demo-2020-02-12-part5

## Bundle

[a8c-motion-background.zip](https://github.com/Automattic/block-experiments/files/4230751/a8c-motion-background.zip) (2020-02-20)

[a8c-motion-background.zip](https://github.com/Automattic/block-experiments/files/4313833/a8c-motion-background.zip) (2020-03-10)


ajlende:
I was just working on GIFs before the meeting. They'll be ready shortly 🙂

blocks/motion-background/motion-background.js
```diff
@@ -0,0 +1,351 @@
...
+	const globalState = {
+		time: window.performance.now(),
+		mouse: [ 0, 0 ],
+	};
+
+	/**
+	 * Hack to move the background behind the canvas since the canvas is added
+	 * underneath the editor to allow for UI elements to draw overtop.
+	 */
+	function replaceBackground() {
```
ajlende: @jasmussen This is one of the "hacky things" I had to do to get layering to look good.

In order to allow for more than 2-32 blocks (depending on the device/browser/heuristics they use for the maximum number of webgl contexts), I have a [single canvas for all the blocks](https://webglfundamentals.org/webgl/lessons/webgl-multiple-views.html). In order for text and UI elements to draw on top of it, the canvas is placed underneath the `#editor` div. However, themes like Twentytwenty have a background color applied to the `.editor-styles-wrapper` div which is rendered by React. Since React owns it and I don't have access to React from here, this is the best I could think of.

One possibility if we want to change the editor might be to have a place for things like this in the React DOM that components can render into with a [portal](https://reactjs.org/docs/portals.html); although, I don't know how practical that would be.

jasmussen:
Took this for a spin today. 

REALLY good work. While there are a few ifs and buts, you've put this together exceptionally well, with great taste for the overall functionality. Nice work.

A couple of things. I get a gray background:

Screenshot 2020-02-24 at 15 27 19

This is in a G2 branch, which may be the cause of it, but a color is set on the `editor-styles-wrapper` element:

Screenshot 2020-02-24 at 15 30 30

The initial state was missing some plasma for me:

Screenshot 2020-02-24 at 15 28 06

It seems like that's because it's missing some gradients?

Screenshot 2020-02-24 at 15 28 28

It also did not work with just colors:

Screenshot 2020-02-24 at 15 28 40

I also couldn't get it to work with an Image. But I suspect it's possibly my local dev environment that's throwing a wrench in the wheels, but I'm sure you'll let me know.

Other than that, we'll want a new block icon — I can provide one, unless Pablo wants to jump in. 

As a followup or future iteration, it would be interesting to explore moving some of those sidebar options into a button on the toolbar. I imagine a button that holds a gradient swatch, and when you click it you get either a modal dialog or a dropdown, whichever holds the content, and you can pick your plasma from there. Something to think about.

blocks/motion-background/motion-background.js
jasmussen: Ah, I suspect this is what's causing the gray background I just mentioned. That was in TwentyTwenty.

I've no problem with hacks per se, but we'll probably want to find a workaround here, as `editor-styles-wrapper` is the equivalent of the document root for the theme. It is a class that exists to attach editor styles to, so it is expected for themes that style the editor to put background colors and other things here. [Here's more](https://developer.wordpress.org/block-editor/developers/themes/theme-support/#editor-styles), but the gist is that if you load `editor-style.scss` in using `add_editor_style`, the following:

```
body {
background: red;
}
```

gets _rewritten_ / transcoded into the following:

```
.editor-styles-wrapper {
background: red;
}
```

Is there an upstream patch we can consider to enable webGL blocks like this?

ajlende:
@jasmussen The gray background is an artifact of G2. The block should be working properly _behind_ that background, so for now you can just remove the gray background to see the animated background. I'll have to update my background hack when G2 lands.

```css
.block-editor-editor-skeleton__content {
	background: transparent !important;
}
```

ajlende:
And I agree about moving the gradient options to a toolbar. The gradient that's getting mixed isn't actually a liner gradient which is why I didn't want to use the existing gradient picker. Instead it's 4 colors in the corners of a rectangle that are blended with gamma so the colors don't get muddy. Having a preview of the un-mixed gradient might be helpful, but it's going to require some creative workarounds for drawing on top of the UI too.

un-mixed gradient

ajlende:
@pablohoneyhoney I got it working again after the changes from WordPress/gutenberg#19082 broke my background hack. The fix I merged into this branch is the one that doesn't require changes to Gutenberg (i.e. WordPress/gutenberg#20483).

The UX is a little weird because the only way I could layer a canvas underneath the UI was from within a block. So now you have a canvas provider block that allows the motion-background as an inner block.

And the code is a little weird because I have to calculate positioning slightly differently in the editor which required a bunch of code duplication.

I still think Gutenberg requires some changes to overcome the awkward UX—we can discuss them on WordPress/gutenberg#20483—but I'm going to squash and merge this for now to document an approach that works prior to any Gutenberg changes that might happen. We can also continue to iterate on the UI and default values in another PR.

PR #22, opened 2020-01-20
Fix grid block

jasmussen:
In recent releases of the plugin, the Grid block has broken in a few ways:

- The placeholder is no longer full width in latest plugin
- The placeholder looks off
- The overlay disappeared

This PR fixes all that.

There's some back compat CSS kept here, but with comments now. It fixes this:

- Placeholder is full width
- Overlay grid is visible again
- Individual columns can be resized.

Before:

Screenshot 2020-01-20 at 11 37 54

After:

Screenshot 2020-01-20 at 11 38 01

Screenshot 2020-01-20 at 11 31 29

Screenshot 2020-01-20 at 12 46 03

### DIFF
--- a/blocks/layout-grid/editor.scss
+++ b/blocks/layout-grid/editor.scss
@@ -16,6 +16,10 @@
 	margin-bottom: 0;
 }
 
+[data-type="jetpack/layout-grid"] > div {
+	width: 100%;
+}
+
 [data-type="jetpack/layout-grid-column"].wp-block { // Selector needs specificity to override.
 	// 2. Unset margins and paddings for column container.
 	padding-left: 0;
@@ -39,6 +43,7 @@
 
 [data-type="jetpack/layout-grid"] {
 	// Pending refinements, hide the v6.3 dashed outlines on grid children, they aren't helpful here.
+	// @todo Revisit/remove for 5.4.
 	.block-editor-block-list__block:not(.is-selected) > .block-editor-block-list__block-edit::before {
 		border-color: transparent !important;
 	}
@@ -89,7 +94,7 @@
 // Hide the hover style & breadcrumbs for individual column containers.
 // This helps simplify the experience for resizing columns.
 // The hover effect may be removed entirely upstream, in which case we can remove this.
-// Ticket to watch: https://github.com/WordPress/gutenberg/issues/17088
+// @todo Revisit/remove for 5.4.
 [data-type="jetpack/layout-grid-column"].is-hovered > .block-editor-block-list__block-edit {
 	// Don't show breadcrumbs.
 	> .block-editor-block-list__breadcrumb {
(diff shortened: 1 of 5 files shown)


johngodley:
@jasmussen is the `IconButton`/`Button` change dependant on a specific version of Gutenberg? I think I tried a similar change elsewhere and on one site with an older version of Gutenberg the change to `Button` didn't work.

jasmussen:
Excellent question. I just saw the deprecation notice in the console and figured I'd make the change. I can revert that if we want to maximize compatibility?

johngodley:
I think it might be necessary, yeah. Here's what the layout selection looks with Gutenberg from sometime in December last year (not sure of exact version):

Create_blank___Pages_by_WordPress_com

After updating to Gutenberg 7.2 it becomes:

Add_New_Post_‹_Test_Tinker_site_—_WordPress

Is the move from centred aligned to left aligned part of the change too?

johngodley:
Everything seems to work great otherwise, using 7.2. With Twenty Twenty as the theme I noticed there's a double-border and grey placeholder in the corners thing going on:

Add_New_Post_‹_Test_Tinker_site_—_WordPress

Not sure if important or expected.

There are some problems with the current version of Gutenberg in WordPress 5.3 though (around Gutenberg 6.5 ish).

There's a giant ugly border when a placeholder is clicked:

Edit_Post_‹_Test_Tinker_site_—_WordPress

The right hand border disappears when the grid block is selected:

Edit_Post_‹_Test_Tinker_site_—_WordPress

Also there's not a lot of vertical space between grid blocks. Here are two grid blocks. Selecting one highlights the block right up to the placeholders of the previous block:

Edit_Post_‹_Test_Tinker_site_—_WordPress

Contrast this to 7.2 where there is a ton of space:

Edit_Post_‹_Test_Tinker_site_—_WordPress

Again, I'm not sure if this is a design change in Gutenberg itself.

jasmussen:
Thanks for incredible testing. I think I'll give this a round of polish to address the points.

jasmussen:
Restored the icon button and fixed the placeholder:

Screenshot 2020-01-21 at 09 46 53
Screenshot 2020-01-21 at 09 47 08

The left aligned placeholder content is intentional, but in need of more polish.

jasmussen:
> With Twenty Twenty as the theme I noticed there's a double-border and grey placeholder in the corners thing going on:

Not really intentional, but also an upstream issue. Also one that will be addressed with the G2 UI.

> There's a giant ugly border when a placeholder is clicked:

Good catch. I made this better:

Screenshot 2020-01-21 at 09 54 48

But it's probably not worth optimizing that much further, as a lot of the double borders are refactored away in subsequent versions.

> The right hand border disappears when the grid block is selected:

Fixed.

> Also there's not a lot of vertical space between grid blocks. Here are two grid blocks. Selecting one highlights the block right up to the placeholders of the previous block:
> Contrast this to 7.2 where there is a ton of space:

Yeeerp, nice catch. Should be fixed.

johngodley:
Confirming that the placeholder is fixed, as is the right border disappearing. The small spacing between two grid blocks in WP 5.3 is still there though.

Edit_Post_‹_Test_Tinker_site_—_WordPress

jasmussen:
I looked at that, and I think we should keep it as is. I believe the gray appender has no margin on its own in earlier versions, but it does in later versions. And for grids with actual content, the content inside will have its margins space things out. So I suppose it's sort of accurate.

johngodley:
Alright, let's go with this then!

jasmussen:
Thank you!

PR #23, opened 2020-01-23
Polish grid further

jasmussen:
Alright, this fixes the two issues uncovered yesterday, plus another one:

- Margins were too wide
- Resize handles were visible on column(s) when they shouldn't be
- Also, just discovered, the grid wasn't visible when resizing column(s)

### DIFF
--- a/blocks/layout-grid/src/grid-resize.scss
+++ b/blocks/layout-grid/src/grid-resize.scss
@@ -31,7 +31,7 @@
 
 	// Every block, including nested blocks, are born with an intrinsic minimum margin.
 	// This resets that for the containers, which aren't meant to have that intrinsically.
-	> .wp-block > .editor-block-list__block-edit > [data-block] {
+	> .wp-block > .block-editor-block-list__block-edit > [data-block] {
 		margin-top: 0;
 		margin-bottom: 0;
 	}
--- a/blocks/layout-grid/src/grid.scss
+++ b/blocks/layout-grid/src/grid.scss
@@ -18,7 +18,7 @@
 
 	// Every block, including nested blocks, are born with an intrinsic minimum margin.
 	// This resets that for the containers, which aren't meant to have that intrinsically.
-	> .wp-block > .editor-block-list__block-edit > [data-block] {
+	> .wp-block > .block-editor-block-list__block-edit > [data-block] {
 		margin-top: 0;
 		margin-bottom: 0;
 	}
@@ -161,7 +161,7 @@ $light-opacity-light-800: rgba(#fff, 0.45);
 	}
 }
 
-/* When clicked, hide the fake drag handles and show the real one */
+// When clicked, hide the fake drag handles and show the real one.
 .wp-block-jetpack-layout-grid__resizing {
 	user-select: none;
 	z-index: 10000;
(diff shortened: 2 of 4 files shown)


blocks/layout-grid/src/grid-resize.scss
```diff
@@ -31,7 +31,7 @@
 
 	// Every block, including nested blocks, are born with an intrinsic minimum margin.
 	// This resets that for the containers, which aren't meant to have that intrinsically.
-	> .wp-block > .editor-block-list__block-edit > [data-block] {
+	> .wp-block > .block-editor-block-list__block-edit > [data-block] {
```
jasmussen: `.editor` as a prefix is deprecated in favor of `.block-editor`, which should have been around for a while.

blocks/layout-grid/src/grid.scss
```diff
@@ -18,7 +18,7 @@
 
 	// Every block, including nested blocks, are born with an intrinsic minimum margin.
 	// This resets that for the containers, which aren't meant to have that intrinsically.
-	> .wp-block > .editor-block-list__block-edit > [data-block] {
+	> .wp-block > .block-editor-block-list__block-edit > [data-block] {
```
jasmussen: I think this is dead CSS, but I can't be sure. I searched the DOM for `wpcom-resize-grid` but can't find it. What's it used for?

blocks/layout-grid/src/grid.scss
```diff
@@ -127,7 +127,7 @@
 }
 
 [data-type="jetpack/layout-grid"].is-selected .wp-block-jetpack-layout-resizable .wp-blocks-jetpack-layout-grid__resize-handles .components-resizable-box__handle,
-[data-type="jetpack/layout-grid"] [data-type="jetpack/layout-grid-column"].is-selected .wp-blocks-jetpack-layout-grid__resize-handles .components-resizable-box__handle {
+[data-type="jetpack/layout-grid"] .wp-block-jetpack-layout-resizable [data-type="jetpack/layout-grid-column"].is-selected .wp-blocks-jetpack-layout-grid__resize-handles .components-resizable-box__handle {
```
jasmussen: The addition of this one ensures the handles are only visible when resizable. My bad.

blocks/layout-grid/src/grid.scss
```diff
@@ -161,7 +161,7 @@ $light-opacity-light-800: rgba(#fff, 0.45);
 	}
 }
 
-/* When clicked, hide the fake drag handles and show the real one */
+// When clicked, hide the fake drag handles and show the real one.
```
jasmussen: Yeah I know. But when they're `//` they are stripped from the rendered CSS, whereas `/* */` is kept.

blocks/layout-grid/src/grid.scss
johngodley: I think it is dead, yes

johngodley:
Some of these may be too subtle for my eyes! Saying that, going through your points:

- The grid lines are vertically as tall as they were before. I don't know if this is the margin issue?
- Resize handles seem to appear in the right place, even on small width screens
- The grid is visible when resizing

There's one thing with a non-Gutenberg theme enabled:

Edit_Post_‹_Test_Tinker_site_—_WordPress

The `li` items for each choice have a bullet style applied.

Otherwise, everything seems 👍

jasmussen:
I can't reproduce the list issue, but I think the CSS i just pushed should fix it.

johngodley:
That did the trick! Let's go with this

PR #34, opened 2020-03-03
Layout grid: fix long lines not wrapping

johngodley:
Long words on themes without specific long word wrapping CSS can break out of the layout grid.

Layout Grid Test

This adds a specific word wrap to grid columns to make sure a long word doesn’t break out.


jasmussen:
Thank you John.

PR #35, opened 2020-03-04
Add a ‘remove end gutters’ option

### LINKED ISSUE #29: Layout Grid: Option to remove gutters on outside
Having gutters on outside works most of the time, but sometimes, it'd be handy to have an option to remove the gutters on the outside.

For example, when a Layout Grid block is inside of a Group block with a background color, it ends up having extra side paddings because of the paddings from the Group block.

Also, you might want the images in the columns to stick to the browser edge...

What do you think?

Artboard

johngodley:
This adds an option to remove the end gutters from the layout grid. The option is available in the parent grid:

Edit_Post_‹_Latest_—_WordPress

Enabling the option removes gutters in the editor.

Before:

Edit_Post_‹_Latest_—_WordPress

After:

Edit_Post_‹_Latest_—_WordPress

The same applies to the front end.

Fixes #29

### DIFF
--- a/blocks/layout-grid/src/grid-overlay.scss
+++ b/blocks/layout-grid/src/grid-overlay.scss
@@ -48,6 +48,10 @@
 	}
 }
 
+.wp-block-jetpack-layout-grid__nowrap .wpcom-overlay-grid {
+	padding-left: 0;
+	padding-right: 0;
+}
 
 body.is-resizing .wpcom-overlay-grid .wpcom-overlay-grid__column {
 	border-left: 1px solid #e2e4e7;
--- a/blocks/layout-grid/src/grid-overlay.scss
+++ b/blocks/layout-grid/src/grid-overlay.scss
@@ -48,7 +48,7 @@
 	}
 }
 
-.wp-block-jetpack-layout-grid__nowrap .wpcom-overlay-grid {
+.wp-block-jetpack-layout-gutter__nowrap .wpcom-overlay-grid {
 	padding-left: 0;
 	padding-right: 0;
 }
--- a/blocks/layout-grid/src/grid/css-classname.js
+++ b/blocks/layout-grid/src/grid/css-classname.js
@@ -125,7 +125,7 @@ export function getAsCSS( columns, attributes = {} ) {
 	}
 
 	if ( ! attributes.addGutterEnds ) {
-		classes[ 'wp-block-jetpack-layout-grid__nowrap' ] = true;
+		classes[ 'wp-block-jetpack-layout-gutter__nowrap' ] = true;
 	}
 
 	return classes;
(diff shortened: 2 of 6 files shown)


johngodley:
@jasmussen appreciate any feedback here. I've gone for a negative option (i.e. you enable the option to disable something) due to existing blocks not having the feature, but the logic could be inverted if that's better.

jasmussen:
Here's what I'm seeing:

gutters

This seems generally good, except for a few things.

First off the behavior — "Remove end gutters" off by default, or "Add end gutters" on by default? Right at this moment, I'm leaning towards the latter.

Let's go with a switch, btw, instead of a checkbox, as the visual effect is instant. So:

```
[switch] Add end gutters
Toggle to remove the spacing left and right of the grid.
```

I don't love the verbiage, but it's probably fine for a start.

There's an issue with the alignment, though, I think we need to remove the end gutters differently. It looks like the toggle correctly removes the end gutters from the actual _content_, but it doesn't remove the end gutters from the `.wpcom-overlay-grid`. Here's a snippet:

```
.wpcom-overlay-grid {
	// The following rules are for Safari.
	// A CSS grid rendering issue means the child nodes of this grid
	// won't stretch to fit the vertical space, unless the position
	// property is changed from absolute, to something else, and back.
	// By being relative when unselected and absolute when selected, we
	// constantly "refresh" this hack.
	position: relative;
	.is-hovered &,
	.is-selected &,
	.has-child-selected & {
		position: absolute;
	}

	top: 0;
	left: 0;
	bottom: 0;
	right: 0;
	display: grid;
	grid-gap: 24px;
	grid-template-columns: $grid-desktop;

	// This padding adds end gutters.
	padding-left: 24px;
	padding-right: 24px;
```

If you toggle out those two paddings at the end at the same time as for the content, this PR should be good to go!

johngodley:
> It looks like the toggle correctly removes the end gutters from the actual content, but it doesn't remove the end gutters from the .wpcom-overlay-grid

Yep, good point - missed that. Added

> Let's go with a switch

I've switched to `ToggleControl`, if that's what you mean:

Edit_Post_‹_Latest_—_WordPress

The logic is now inverted

jasmussen:
> The logic is now inverted

Looks good, but I feel like it should be enabled by default.

johngodley:
It is enabled by default. The screenshot is just whatever state the block was in when I took it

jasmussen:
Ship it!

jasmussen:
Hmm actually we may want conditional help text.

johngodley:
Alright, message also toggles:

Edit_Post_‹_Latest_—_WordPress

Edit_Post_‹_Latest_—_WordPress

jasmussen:
Okay now ship it. And thank you!

PR #36, opened 2020-03-05
Add configurable gutter sizes to layout grid

### LINKED ISSUE #30: Layout Grid: Setting options for gutters
Currently, the gutter is fixed to 24px, but It'd be great if the user could increase/decrease the gutter size. If the setting allows them to set to `0`, they could even have a gutterless grid that would be handy for media-oriented page layouts.

johngodley:
Adds an option to adjust the layout grid gutter size:

Edit_Post_‹_Latest_—_WordPress

This affects the size of the gap between grid columns.

Huge:

Edit_Post_‹_Latest_—_WordPress

Medium:

Edit_Post_‹_Latest_—_WordPress

Note that the left and right padding is also adjusted.

Fixes #30

### DIFF
--- a/blocks/layout-grid/constants.scss
+++ b/blocks/layout-grid/constants.scss
@@ -4,6 +4,11 @@ $grid-tablet: repeat(8, 1fr);
 $grid-mobile: repeat(4, 1fr);
 
 $grid-gutter: 24px;
+$grid-gutter-none: 0px;
+$grid-gutter-small: 8px;
+$grid-gutter-medium: 16px;
+$grid-gutter-large: $grid-gutter;
+$grid-gutter-huge: 48px;
 $grid-gutter__background-offset: $grid-gutter / 2 + 1px;
 $padding-none: 0px;
 $padding-small: 8px;
--- a/blocks/layout-grid/front.css
+++ b/blocks/layout-grid/front.css
@@ -1032,6 +1032,16 @@
     word-break: break-word;
     word-wrap: break-word; }
 
+  .wp-block-jetpack-layout-grid.wp-block-jetpack-layout-gutter__none {
+    grid-gap: 0px; }
+  .wp-block-jetpack-layout-grid.wp-block-jetpack-layout-gutter__small {
+    grid-gap: 8px; }
+  .wp-block-jetpack-layout-grid.wp-block-jetpack-layout-gutter__medium {
+    grid-gap: 16px; }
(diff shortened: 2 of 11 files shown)


johngodley:
There are two issues, although I'm not sure how important they are.

1) Gutter size 'none' combined with removing end gutters makes no sense. The 'remove end gutters' option is useful for all other sizes, but has no effect with a gutter size of 'none' . Could we hide the option if 'none' is selected?

2) The drag handles can overlap on small gutter sizes:

Edit_Post_‹_Latest_—_WordPress

  It's still functional, and you can select a column and resize just that.  Is this a problem? Can we make the grid gap bigger in the editor? (i.e. a small editor grid gap is the chosen size + 10px or something)

@jasmussen this required surprisingly little work, which makes me suspect there's an underlying problem! Thoughts appreciated. If anything here looks problematic then I think it can be put on the back-burner.

jasmussen:
> Gutter size 'none' combined with removing end gutters makes no sense. The 'remove end gutters' option is useful for all other sizes, but has no effect with a gutter size of 'none' . Could we hide the option if 'none' is selected?

Can you gray it?

> It's still functional, and you can select a column and resize just that. Is this a problem? Can we make the grid gap bigger in the editor? (i.e. a small editor grid gap is the chosen size + 10px or something)

Hmm. it's something to think about — I think it's shippable in that you can select an individual column and resize it. But yes, either making the grid-cap increase when the block is selected (perhaps with a blue background to indicate it's ephemeral somehow) might work. Or we could redesign the resize handles to sit on the "inside" of the block. 

Any thoughts on that? I definitely think we can ship these improvements and iterate the UI further on the side.

johngodley:
> Can you gray it?

Sadly there's no `disabled` option on the `ToggleControl` component. We can probably use CSS to force grey it, and ignore change events, but the control isn't the most straightforward. For now I've removed the option if 'none' is selected. 

If removing the option is ok then yes, I think it's fine to merge, and we can worry about the drag handles later.

jasmussen:
Are you sure? I believe there's supposed to be. Or maybe I'm misremembering and I saw it disabled by wrapping it in a `<Disabled>` container.

johngodley:
Yep, there's no option in `ToggleControl` itself.

> I saw it disabled by wrapping it in a <Disabled> container.

Alright, that somehow magically works. Never seen that component before!

PR #37, opened 2020-03-05
Add IE prefixes to layout grid

### LINKED ISSUE #26: Layout grid: add css prefixes
Add css prefixes for MS-Edge and MSIE-11 support

johngodley:
This uses autoprefixer in conjunction with its [`no-autoplace`](https://github.com/postcss/autoprefixer#control-comments) feature to add IE11/Edge specific prefixes to the grid CSS.

The prefixed CSS does not appear to affect other browsers.

Note that IE11 doesn't support certain grid features, even with prefixes. Specifically `grid-gap` is not available and so the columns have no gaps between them. We could probably fake it with margins/padding in some way, but I'm not sure if it's worth it.

I believe Edge supports `grid-gap`, but since it's now turned into Chrome I've been unable to test.

## Testing

1) Create a post with a grid
2) Without this PR applied, view the post and 'save page as' a complete web page.
3) Start IE, open inspector, and ensure emulation document mode is set to IE11
4) Drop the saved HTML file into IE and confirm that the grid is displayed vertically

before

5) Repeat but with this PR applied. Confirm that the grid is now horizontal
after

Fixes #26

### DIFF
--- a/blocks/layout-grid/front.css
+++ b/blocks/layout-grid/front.css
@@ -5,688 +5,1029 @@
 /**
  * Responsive Grid Options
  */
+/* autoprefixer grid: no-autoplace */
 .wp-block-jetpack-layout-grid {
+  display: -ms-grid;
   display: grid;
   grid-gap: 24px;
+  -ms-grid-columns: (1fr)[4];
   grid-template-columns: repeat(4, 1fr); }
   .wp-block-jetpack-layout-grid.column1-mobile-grid__start-1 > .wp-block-jetpack-layout-grid-column:nth-child(1) {
-    grid-column-start: 1; }
+    -ms-grid-column: 1;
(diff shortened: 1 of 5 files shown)


johngodley:
@jasmussen sorry to keep throwing these at you. This is a fairly straightforward patch that makes things look better, but not perfect, on IE11 (and maybe old Edge). I don't know where we stand in terms of compatibility, especially now that new Edge has turned into Chrome, so I'll defer to you whether this is worth including.

jasmussen:
I'm unable to test this in IE right at this moment. But a) I trust your judgement here, so ship it, we can always improve. And b) we should not bend over backwards to IE11 — I don't think we even explicitly need to support it. If it loads break the site, that seems okay to me.

PR #38, opened 2020-03-05
Firefox fix for a cover block inside the grid block

### LINKED ISSUE #27: Firefox and CSS grid and cover block
There's a difference between how Chrome/Safari and Firefox displays a cover block inside a grid block. 

This is based on an issue reported on [this page](https://amwpapptesting.wordpress.com/tiled-gallery-and-other-testing/) where the bottom black box leaks outside the post in Firefox. I think I've narrowed the complicated page down to this codepen:

https://codepen.io/johngodley/pen/vYOYxJe

In Safari/Chrome the black box is about the same height as the yellow one. In Firefox the black box is a lot bigger. 

The issue seems to be caused by `wp-block-cover` having a `height: 100%` when it follows something else that also has height. That seems to stretch the cover, and grid. Without that the height drops to the `min-height: 430px`.

Is there anything we can do to the grid block to prevent this?

johngodley:
Firefox exhibits different grid behaviour to other browsers when a cover block follows another block inside the same grid column.

Here is an example two column grid with the first column containing an image and a cover block:

Add_New_Post_‹_Latest_—_WordPress

When viewed in Chrome and Safari this displays correctly:

firefox_–_Latest

When viewed in Firefox the cover block breaks out of the grid:

firefox_–_Latest_-_Mozilla_Firefox

This PR applies a a `max-height: 0` which fixes the problem (props to @jasmussen). The fix is specific to Firefox and the cover block. There may be other blocks that cause a similar issue, and they will need to be 'fixed' on an individual basis.

If the behaviour of Firefox changes in the future we can think about removing this fix. 

## Testing

1) Create a two column grid block. In the first column add an image and a cover block (with image)
2) View the post in Firefox and note that it breaks out of the bottom of the post
3) Apply PR
4) View the post in Firefox again and note that the cover is contained

Fixes #27

### DIFF
--- a/blocks/layout-grid/front.css
+++ b/blocks/layout-grid/front.css
@@ -690,3 +690,7 @@
   .wp-block-jetpack-layout-grid .wp-block-jetpack-layout-grid-column * {
     word-break: break-word;
     word-wrap: break-word; }
+
+@-moz-document url-prefix() {
+  .wp-block-jetpack-layout-grid .wp-block-cover {
+    max-height: 0; } }
--- a/blocks/layout-grid/front.scss
+++ b/blocks/layout-grid/front.scss
@@ -106,4 +106,11 @@
 		word-break: break-word;
 		word-wrap: break-word;
 	}
+
+	// Fixes an issue in Firefox where a block in the grid with 100% height causes the grid to increase in height
+	@-moz-document url-prefix() {
+		.wp-block-cover {
+			max-height: 0;
+		}
+	}
 }


johngodley:
@jasmussen alright, final one for the moment I promise! Fix works well. I've tried to restrict impact by making the fix apply only to Firefox, with what I'm lead to believe is the best method of doing that.

jasmussen:
Code looks good. I wish we better knew what was causing this; Firefox is not IE, so there's a chance it's doing the right thing and not Chrome. 

We should also consider an upstream PR to remove the rule from the Cover block in the first place, if it truly isn't needed.

But as an interim fix, and because it's for an edgecase, and scoped to just a Firefox, this seems good. 👍

PR #39, opened 2020-03-05
Add text domains

akirk:
For loading translations of Gutenberg blocks, [we need to specify the text domain](https://developer.wordpress.org/block-editor/developers/internationalization/) with the WordPress.org plugin slug being the correct domain.

This adds the text domain for the following plugins we have on .org:
- [Bauhaus Centenary](https://wordpress.org/plugins/bauhaus-centenary/)
- [Layout Grid](https://wordpress.org/plugins/layout-grid/)
- [Starscape](https://wordpress.org/plugins/starscape/)

This also adds the missing `wp_set_script_translations()` calls to actually load the translations from the language pack.

### DIFF
--- a/blocks/bauhaus-centenary/index.php
+++ b/blocks/bauhaus-centenary/index.php
@@ -15,4 +15,6 @@ add_action( 'init', function() {
 			'editor_style'    => 'wpcom-blocks-editor',
 		]
 	);
+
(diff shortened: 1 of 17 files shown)


johngodley:
Works well, merging.

akirk:
Thanks!

PR #54, opened 2020-03-16
Add Image Compare Block

mkaz:
Adds a new Image Compare Block that allows for placing two images side-by-side (horizontal or vertical) and use a slider to do a before and after image comparison.

Uses code from: https://github.com/NUKnightLab/juxtapose licensed under MPLv2

Example:

image


blocks/juxtapose/src/edit.js
```diff
@@ -28,53 +29,57 @@ const edit= ({ attributes, setAttributes, className }) => {
...
-					value={caption}
-				/>
+								allowedTypes={ [ 'image' ] }
+								labels={ { title: 'Image After' } }
+							/>
+						</div>
+					) }
+				</Placeholder>
 			</div>
+			<TextControl
```
jasmussen: I tried making this a `RichText` component, that was only visible when the block was `isSelected`, but failed. Feel free to take a stab. As you can see I added an `inlineToolbar` prop for bold, etc.

jasmussen:
Pushed some in-progress work:

Screenshot 2020-03-16 at 20 26 14

This isn't how it should look, but it takes a small step towards it. Your idea of two placeholders side by side is interesting, I think we should keep that general idea. But it should be inside a single `<Placeholder>` component. So we need to extract the upload/browse media features from the MediaPlaceholder, and duplicate them twice inside. I can then lay things out. 

Reason being: each placeholder is individually responsive through voodoo. And if we include two side by side, that responsiveness breaks. Easiest way to test is to create a columns block and insert one in each.

blocks/juxtapose/src/edit.js
mkaz: I pushed an update to switch caption to RichText 👍

mkaz:
One feature we may want to add is a starting position. Move the slider to a specific spot within the editor and when the post loads the slider starts there. Right now it defaults to center.

jasmussen:
> Move the slider to a specific spot within the editor and when the post loads the slider starts there. Right now it defaults to center.

If this is more than 5 minutes (literally) of work, I'd call that a really awesome V2 feature we can add :)

pablohoneyhoney:
Here a few more UI options. Any could also have a subtle reaction to hover and drag.

mkaz:
@jasmussen I wasn't quite sure what you wanted to do with the MediaPlaceholder, so with the last change I just pushed up, I swapped out the import with a copy in `upload-placeholder.js` \in our repo.

This will allow you to do any customization we want.

jasmussen:
I'll try and take a look today! I like the tall vertical handle Pablo made.

jasmussen:
Since the last fixes, I cannot get the block to show up in the block library. I get no JS errors, so I don't know what may have happened, probably something in the rename? Marcus does it show up for you in the block library?

blocks/image-compare/editor.scss
```diff
@@ -0,0 +1,30 @@
+
+/* main block wrapper */
```
jasmussen: I have perhaps been exposed too much to the strict Gutenberg principles we have here. Which is: Sentence case and end with a period. And use /* */ if you want the comment to be visible/baked into the CSS, and use // if you are just commenting for someone reading the SCSS file.

blocks/image-compare/editor.scss
```diff
@@ -0,0 +1,30 @@
...
+	max-width: 50%;
+	display: block;
+	text-align: center;
+}
+
+.img-edit-before {
+	border-right: 1px solid #999999;
+}
+
+.caption {
```
jasmussen: This is probably now dead CSS. 

We should get into the habit of a nice prefix on a per-block basis. I.e. for Layout Grid it would be `.layout-grid__caption-style`. In this case, it would be `.image-compare__caption-style`. Outside of being more readable code, it ensures we don't accidentally use a CSS class that a ton of other random things like WordPress themes use, of which there would be a high likelyhood here.

blocks/image-compare/assets/juxtapose.css
```diff
@@ -0,0 +1,285 @@
...
+
+div.jx-control {
+	height: 100%;
+	margin-right: auto;
+	margin-left: auto;
+	width: 3px;
+	background-color: currentColor;
+}
+
+.vertical div.jx-control {
```
jasmussen: I imagine this `.vertical` class comes from the Juxtapose library and so we probably can't/don't want to customize it. But ideally this would be prefixed.

blocks/image-compare/editor.scss
```diff
@@ -0,0 +1,30 @@
...
+	text-align: center;
+}
+
+.caption input[type="text"],
+.caption input[type="text"]:focus {
+	width: 50%;
+	border: none;
+	text-align:center;
```
jasmussen: Space after colon :) — details matter, says your linter friend!

blocks/image-compare/src/edit.js
```diff
@@ -0,0 +1,97 @@
+/**
+ * WordPress dependencies
+ */
+import { InspectorControls, RichText } from '@wordpress/block-editor';
+import { SelectControl, TextControl, Placeholder } from '@wordpress/components';
```
jasmussen: TextControl no onger used.

blocks/image-compare/src/edit.js
```diff
@@ -0,0 +1,97 @@
...
+import { __ } from '@wordpress/i18n';
+
+/**
+ * Internal dependencies
+ */
+import UploadPlaceholder from './upload-placeholder';
+
+const edit = ( { attributes, className, isSelected, setAttributes } ) => {
```
jasmussen: className isn't used.

blocks/image-compare/src/edit.js
```diff
@@ -0,0 +1,97 @@
...
+							juxtapose.scanPage();
+						}, 100 );
+					} }
+				/>
+			</InspectorControls>
+			<div className={ cls } data-mode={ orientation }>
+
+				<Placeholder>
+					{ imageBefore ? (
+						<img src={ imageBefore } />
```
jasmussen: We need an alt attribute. If it can't be dynamic, "First image" or something in that vein is fine.

blocks/image-compare/src/edit.js
```diff
@@ -0,0 +1,97 @@
...
+
+	// if both are defined, add juxtaspose class
+	// the juxtapose library uses class when page is scanned
+	// to find the images and apply the side-by-side magic
+	const cls = ( imageBefore && imageAfter ) ? 'juxtapose' : 'controls';
+
+	return (
+		<>
+			<InspectorControls key="controls">
+				<h4> Orientation </h4>
```
jasmussen: Extra spaces :) 

Also, text needs to be translatable.

blocks/image-compare/src/edit.js
```diff
@@ -0,0 +1,97 @@
...
+				<SelectControl
+					value={ orientation }
+					options={ [
+						{ label: 'Side by Side', value: 'horizontal' },
```
jasmussen: Needs to be translatable.

blocks/image-compare/src/edit.js
```diff
@@ -0,0 +1,97 @@
...
+						<div className="img-edit-before">
+							<UploadPlaceholder
+								onSelect ={
+									( el ) => {
+										setAttributes( { imageBefore: el.url } );
+										juxtapose.scanPage();
+									}
+								}
+								allowedTypes={ [ 'image' ] }
+								labels={ { title: 'Image Before' } }
```
jasmussen: Translatable.

blocks/image-compare/src/edit.js
```diff
@@ -0,0 +1,97 @@
...
+				<RichText
+					tagName="figcaption"
+					placeholder={ __( 'Write caption' ) }
+					value={ caption }
+					onChange={ ( value ) =>
+						setAttributes( { caption: value } )
+					}
+					// these are from image caption, do we need?
+					//unstableOnFocus={ this.onFocusCaption }
+					//isSelected={ this.state.captionFocused }
```
jasmussen: We probably do. The `isSelected` I couldn't easily make work. It's there to "pop out" the caption field when the block is selected, but hide it otherwise.

blocks/image-compare/src/index.js
```diff
@@ -0,0 +1,73 @@
...
+ */
+import edit from './edit';
+
+export function registerBlock() {
+
+	registerBlockType( 'jetpack/image-compare-block', {
+
+		title: 'Image Compare',
+
+		icon: 'image-flip-horizontal',
```
jasmussen: Let's use this icon:
Screenshot 2020-03-18 at 09 46 31

````
<SVG xmlns="http://www.w3.org/2000/svg" viewBox="0 0 24 24">
	<Path d="M3 5v14c0 1.1.9 2 2 2h8.5V3H5c-1.1 0-2 .9-2 2zm9 14.5H5c-.3 0-.5-.2-.5-.5v-2.4l4.1-3 3 1.9c.1.1.3.1.4.1v3.9zm0-5.7L9 12c-.1-.1-.3-.1-.4-.1-.2 0-.3 0-.4.1l-3.6 2.6V5c0-.3.2-.5.5-.5h7v9.3zM19 3h-4v1.5h4c.3 0 .5.2.5.5v14c0 .3-.2.5-.5.5h-4V21h4c1.1 0 2-.9 2-2V5c0-1.1-.9-2-2-2z" />
</SVG>
```

Remember to import `SVG` and `Path` components so it works in React Native.

I believe you can just drop the SVG right inline here.

blocks/image-compare/src/index.js
```diff
@@ -0,0 +1,73 @@
...
+
+		category: 'layout',
+
+		attributes: {
+			imageBefore: {
+				type: 'string',
+				source: 'attribute',
+				attribute: 'src',
+				selector: '.imgBefore',
```
jasmussen: Do we control this classname? I'd love to avoid camel-casing :)

blocks/image-compare/src/index.js
```diff
@@ -0,0 +1,73 @@
...
+				selector: 'figcaption',
+			},
+			orientation: {
+				type: 'string',
+			}
+		},
+
+		example: {
+			attributes: {
+				imageBefore: 'https://placedog.net/240/120',
```
jasmussen: We need to ensure that we are allowed not only to use these images (license), but that their website allows hotlinking.

Wikimedia does. We could use these two photos of Lynda Carter as Wonder Woman:


https://commons.wikimedia.org/wiki/File:Lynda_Carter_Wonder_Woman_black_and_white.jpg


https://commons.wikimedia.org/wiki/File:Lynda_Carter_Wonder_Woman.JPG

Wikimedia explicitly _allows_ hotlinking, and those two pictures are public domain, so we're as safe as can be.

blocks/image-compare/src/index.js
```diff
@@ -0,0 +1,73 @@
...
+				imageAfter: 'https://placedog.net/239/119',
+				caption: 'Doggos',
```
jasmussen: Needs to be translatable.

blocks/image-compare/src/index.js
```diff
@@ -0,0 +1,73 @@
...
+			},
+		},
+
+		edit,
+
+		save: ({ attributes }) => {
+			return (
+				<>
+					<figure className='juxtapose' data-mode={attributes.orientation}>
```
jasmussen: `"` instead of `'` for classnames.

jasmussen:
I left a metric ton of comments. I would've fixed all these myself if I could've gotten the branch to work properly. When it does, I'd like to polish the placeholder and try and get pablo's drag handle working.

blocks/image-compare/editor.scss
mkaz: Removed.

blocks/image-compare/editor.scss
mkaz: Fixed.

blocks/image-compare/src/edit.js
mkaz: Fixed.

blocks/image-compare/src/edit.js
mkaz: Fixed.

blocks/image-compare/src/edit.js
mkaz: Fixed.

blocks/image-compare/src/edit.js
mkaz: Fixed.

blocks/image-compare/src/edit.js
mkaz: Fixed.

blocks/image-compare/src/index.js
mkaz: Added.

blocks/image-compare/src/index.js
mkaz: Switched.

blocks/image-compare/src/index.js
mkaz: Fixed.

blocks/image-compare/src/index.js
mkaz: Fixed.

jasmussen:
Okay, pushed a ton of work. Placeholder:

Screenshot 2020-03-19 at 09 53 20

Narrow placeholder:

Screenshot 2020-03-19 at 09 54 04

Very narrow placeholder:

Screenshot 2020-03-19 at 09 54 08

Placeholders partially filled in:

Screenshot 2020-03-19 at 09 54 12

The full flow:

image compare

There's only one thing left to do, and that's fix the preview: right now the `juxtapose` JS isn't applied to the preview at all:

Screenshot 2020-03-19 at 09 37 48

Once it is (and if we can't we'll have to do something else), we can ship this!

jasmussen:
Polished a few more things, the drop target, made the selector in the sidebar into a radio group, and got the vertical one working:

vertical

@mkaz outside of the hover preview, do you think you can help me with this JS error/warning that pops up in my console?

```
Warning: React does not recognize the `isSelected` prop on a DOM element. If you intentionally want it to appear in the DOM as a custom attribute, spell it as lowercase `isselected` instead. If you accidentally passed it from a parent component, remove it from the DOM element.
```

jasmussen:
Okay, I pushed some tweaks:

Screenshot 2020-03-19 at 16 02 47

New icons, no border, new verbiage, fixed arrows.

There are two remaining issues that I could use your help with, @mkaz, and I believe that React error I posted above is related to one of them:

1. In changing the inspector control to use two radio buttons instead of the dropdown, I got myself into some trouble with the controls, and it does not appear that you can actually switch between the two anymore. 

2. We need to run the juxtapose JS on the block preview. 

The way I test that hover-only interfaces is this: 

- open the inspector
- use Chrome (or Chromium Edge)
- open the console
- paste in this: `setTimeout(function(){debugger;}, 2000);`
- open the inserter and search for the image compare block so it's ready
- press "Enter" in the console. You now have two seconds to hover the Image Compare block in the library. After the two seconds are passed, the interface is locked and you can now inspect the preview.

blocks/image-compare/src/edit.js
```diff
@@ -0,0 +1,117 @@
...
+ * Internal dependencies
+ */
+import UploadPlaceholder from './upload-placeholder';
+
+/* global juxtapose */
+
+const edit = ( { attributes, isSelected, setAttributes } ) => {
+	const { imageBefore, imageAfter, caption, orientation } = attributes;
+
+	const RadioControlWithState = ( props ) => {
```
jasmussen: I'm pretty sure I messed up this bit, with regards to setting the orientation of the interface. If you can help me get this working, I'd appreciate it.

pablohoneyhoney:
Is it me or the circle does look cut?

jasmussen:
> Is it me or the circle does look cut?

I noticed that the briefest of times. If it is what I think it is, it's because the positioning places the circle on subpixels, but the renderer rounds up to nearest whole pixel, causing a crop. I have two ideas for how to fix that, which I'll try.

blocks/image-compare/src/edit.js
mkaz: Fixed.

jasmussen: You rock. Let the record show I owe you a beverage.

jasmussen:
Okay, I think this one is good. Let's ship it!

PR #60, opened 2020-03-23
Improve/create preview for Image Compare block.

### LINKED ISSUE #55: Image Compare - Improve Block Preview
The preview is not accurate, it shows the two images side-by-side and not a good representation of the block.

image

There isn't a way to trigger an action on Preview, so we can't call the library code that handles.

So we'll need the default state with two images, without calling the JS code, to look proper in Preview.

From chat with Joen:

I think we can and I think I might be able to make that happen — basically what happens is, as soon as you add image number 2, the JS is invoked to make the thing interactive, this replaces the  so what you see in the preview is the placeholder but without that JS

I could very probably make it so that the two images look sort of stacked side by side in a representative preview in the “no js” state, which would work for the preview, and get replaced by the JS when present

jasmussen:
This PR styles the preview of the Image Compare block to better resemble the block itself. Fixes #55.

Screenshot 2020-03-23 at 10 37 34

I did try first a horizontal split, because that's the default. But the two images that we are hotlinking (with permission) from Wikimedia do not align well in that state:

Screenshot 2020-03-23 at 10 34 44

### DIFF
--- a/blocks/image-compare/editor.scss
+++ b/blocks/image-compare/editor.scss
@@ -36,3 +36,48 @@
 [data-type="jetpack/image-compare-block"]:not(.is-selected) .image-compare__comparison {
 	pointer-events: none;
 }
+
+// When juxtapose isn't loaded, we style the block to resemble the end result.
+// This effectively styles the block to look good in the block preview.
+.juxtapose .components-placeholder {
+	border: none;
+	padding: 0;
+	box-shadow: none;
+
+	.components-placeholder__label {
+		display: none;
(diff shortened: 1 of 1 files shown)


jasmussen:
Note that I did not include the drag handle in the center of the white line in the preview, as it would duplicate a bunch of CSS for the preview, and did not seem to add much value. However if you all feel it should be there, I can add it.

mkaz:
Looks good. 👍 
Tested it out and no problems.

PR #61, opened 2020-03-25
Image Compare - fix images when window resized

mkaz:
The photos aren't resized properly when the window is resized.

Screenshot 2020-03-23 at 13 54 12

This PR adds using the `useResizeObserver` hook to trigger calling the optimizeWrapper function.

### DIFF
--- a/blocks/image-compare/src/edit.js
+++ b/blocks/image-compare/src/edit.js
@@ -1,8 +1,15 @@
 /**
+ * External dependencies
+ */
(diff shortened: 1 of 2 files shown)


jasmussen:
This seems fine to me 👍 👍!

jasmussen:
Pulling your latest, this still feels good to me! If you need a code sanity check, feel free to, otherwise, ship it!

PR #71, opened 2020-04-09
Grid block version 2!

jasmussen:
This PR does a number of things to polish the grid block. But right up front, it does kind of change the requirements, but to something I think is reasonable:

- It now requires WordPress 5.4, 
- OR it requires the Gutenberg plugin version 7.7 or higher.

It should _work_ with 5.3, but the layout may not be ideal in all situations. This is speaking of the _editor_ view — the frontend appearance should be unchanged and work wherever. 

Video: https://cloudup.com/c_pvr3hp9oi

It does these things:

- It rethinks the UI in a world where the G2 UI exists. That is — no more unsetting of paddings and margins, but instead embracing the footprint of each block.
- As a result, much hacky code has been removed.
- Also as a result, the grid now no longer "sizes down" when selected, which fixes some regressions we've seen from the hacks that made that work. 
- Polish to the appender, and fixing a typo.

The new approach is much more resilient.

- Drag handles have been redesigned to be simpler and more G2-esque. If this works well, we can look at submitting this to the core project. 
- When the block is full-wide and has end-gutters, no change.
- When the block is full-wide and hides end-gutters, drag handles are "collapsed inwards". This means a drag handle that sits way on the edge of the screen is still fully visible and does not create horizontal scrollbars. 

No endgutters:

Screenshot 2020-04-09 at 11 23 07

Endgutters:

Screenshot 2020-04-09 at 11 22 48

There will be a situation where you are running 5.3 without the plugin and using no endgutters, where some blocks in the leftmost column have their side UI cut off. I think this is an acceptable edgecase to not optimize for anymore.

### DIFF
--- a/blocks/layout-grid/src/index.js
+++ b/blocks/layout-grid/src/index.js
@@ -32,7 +32,7 @@ function getColumnAttributes( total, breakpoints ) {
 export function registerBlock() {
 	registerBlockType( 'jetpack/layout-grid', {
 		title: __( 'Layout Grid', 'layout-grid' ),
-		description: __( 'Align blocks to to a global grid, with support for responsive breakpoints.', 'layout-grid' ),
+		description: __( 'Align blocks to a global grid, with support for responsive breakpoints.', 'layout-grid' ),
 		icon: GridIcon,
 		category: 'layout',
 		supports: {
--- a/blocks/layout-grid/editor.scss
+++ b/blocks/layout-grid/editor.scss
@@ -21,29 +21,12 @@
 	margin: 0;
 	padding-left: 0;
 	padding-right: 0;
-
-	// 3. Unset for contents of container.
-	// @todo Revisit/remove for 5.4.
-	> .block-editor-block-list__block-edit {
-		padding-left: 0;
-		margin-left: 0;
-		padding-right: 0;
-		margin-right: 0;
-	}
 }
 
 
 /**
  * Visual Glitches
  */
-
-[data-type="jetpack/layout-grid"] {
-	// Pending refinements, hide the v6.3 dashed outlines on grid children, they aren't helpful here.
-	// @todo Revisit/remove for 5.4.
-	.block-editor-block-list__block:not(.is-selected) > .block-editor-block-list__block-edit::before {
-		border-color: transparent !important;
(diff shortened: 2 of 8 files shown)


blocks/layout-grid/src/index.js
```diff
@@ -32,7 +32,7 @@ function getColumnAttributes( total, breakpoints ) {
 export function registerBlock() {
 	registerBlockType( 'jetpack/layout-grid', {
 		title: __( 'Layout Grid', 'layout-grid' ),
-		description: __( 'Align blocks to to a global grid, with support for responsive breakpoints.', 'layout-grid' ),
```
johngodley, line 35: Well spotted!

johngodley:
This works great, and looks the part too. Nicely done!

If we're moving to 5.4 then we can swap out all the `<IconButton>` uses for `<Button>`. We can do that in another PR though.

We'll need to do some additional testing on .com, but that can happen after this has merged and when a diff has been prepared.

johngodley:
Oh, sometimes when the handles are close together they kind of merge like this:

image

I'm assuming this is ok

johngodley:
One more! On a mobile display the blue border gets cropped a bit:

image

Again, don't know if it's important.

blocks/layout-grid/src/index.js
jasmussen: Props @donalirl in https://github.com/Automattic/wpcom-blocks/issues/169#issue-597041838!

jasmussen:
Great feedback, all of it. I think it's worth it to move to 5.4, but I'm open to other opinions here. 

Regarding the handles, yes I think it's okay they stack like that. That's actualy one of the reasons for the different drag handle. When there are gutters:

Screenshot 2020-04-09 at 13 33 45

When there are no gutters:

Screenshot 2020-04-09 at 13 33 37

👆that up there is the "right edge" handle. 

For the mobile cropped borders, yes I noticed that too, and submitted https://github.com/WordPress/gutenberg/pull/21498 for discussion around that. It comes with a teensy tradeoff I thin is worth it.

johngodley:
👍 no problem moving to 5.4 - let's merge!

PR #73, opened 2020-04-09
Add new icons.

jasmussen:
This PR adds new icons for the grid block. Block itself:

Screenshot 2020-04-09 at 15 09 42

A column:

Screenshot 2020-04-09 at 15 10 15

For reference, here's the stock Columns icon:

Screenshot 2020-04-09 at 15 04 55

### DIFF
--- a/blocks/layout-grid/constants.scss
+++ b/blocks/layout-grid/constants.scss
@@ -10,11 +10,13 @@ $grid-gutter-medium: 16px;
 $grid-gutter-large: $grid-gutter;
 $grid-gutter-huge: 48px;
 $grid-gutter__background-offset: $grid-gutter / 2 + 1px;
-$padding-none: 0px;
-$padding-small: 8px;
-$padding-medium: 16px;
-$padding-large: 24px;
-$padding-huge: 48px;
+
+// These can be deprecated.
+$padding-none: $grid-gutter-none;
+$padding-small: $grid-gutter-small;
+$padding-medium: $grid-gutter-medium;
+$padding-large: $grid-gutter;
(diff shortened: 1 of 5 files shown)


pablohoneyhoney:
Among the above, I think the second might work best. 

Or try something more specific to _layout_

block-layout

jasmussen:
> Among the above, I think the second might work best.

Can you clarify, do you mean re-using the icon for the Columns block? This one?

Screenshot 2020-04-10 at 07 43 18

I do agree that this block is more about _layout_, and for that reason the icon you suggested is better. However it does feel wrong to me that the icon shows a layout you can't achieve with the block. 

But I don't feel this strongly, so I'm happy to update with that icon, if you prefer it.

jasmussen:
Alright, I updated with a bunch of things, with some feedback from Pablo:

grid

- Square handles
- New shading for the gutters

Pablo also suggested the handles could have a 24px min-height and 48px max-height, but this turned out to be quite the challenge to make work. I believe for now we should go with what we have here, test out the handles, and then based on how they feel we can try and submit the handle improvement as an upstream improvement to the core project itself, and see if we can revisit the min/max-elasticity there. 

The remaining issue is the icon, and the original purpose of the PR here. I keep being a fan of the original one, not because it's very layout-y, but because it does in fact consist of vertical columns like the block supports. Additionally, there are actually _two_ icons — one indicating the layout grid block itself, and another indicating one of the columns inside being selected. Observe:

icons

If we do want to go with a different icon, we need an additional different icon to indicate _layout grid column_.

jasmussen:
I'm merging this as is and promise to revisit based on feedback!

PR #78, opened 2020-04-24
Move the currently selected breakpoint to the toolbar

### LINKED ISSUE #72: Grid Block: Move device indication to toolbar
Instead of this:

Screenshot 2020-04-09 at 14 46 13

Can we do this:

Screenshot 2020-04-09 at 14 45 00

The button in the toolbar would be a dropdown that doubled as the breakpoint selector that sits in the sidebar.

johngodley:
Remove the selected breakpoint from under the grid and move it to the toolbar. It's also changed into a dropdown menu where it can be changed.

image

Fixes #72

### DIFF
--- a/blocks/layout-grid/editor.scss
+++ b/blocks/layout-grid/editor.scss
@@ -133,25 +133,6 @@
 	}
 }
 
-.jetpack-layout-grid-previewing {
-	font-family: -apple-system,BlinkMacSystemFont,"Segoe UI",Roboto,Oxygen-Sans,Ubuntu,Cantarell,"Helvetica Neue",sans-serif;
-	font-size: 13px;
-	position: absolute;
-	bottom: -24px;
-	left: 0;
-	padding: 0 8px;
-	height: 24px;
-	background-color: rgba(0,0,0,0.8);
-	color: #fff;
-	display: none;
-}
-
-[data-type="jetpack/layout-grid"].has-child-selected .jetpack-layout-grid-previewing,
-[data-type="jetpack/layout-grid"].is-selected .jetpack-layout-grid-previewing {
-	display: block;
-}
-
-
 /**
  * Redesign resize handles to handle literal edge cases.
  */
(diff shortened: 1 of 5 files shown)


johngodley:
@jasmussen is this what you had in mind? It seems a good improvement and works well. There's a small amount of wobble when you change the breakpoint and the text in the toolbar changes width, but this would seem to be expected.

jasmussen:
Yes! That's glorious! Taking a look right now.

jasmussen:
I pushed a small change ot make the menu dark.

The wobble is a bit annoying, I don't know why that happens, I think it centers the popover according to the width of the button, because if I fix that width, then it stops wobbling. 

One thing we could do is to change the text labels to icons. The gutenberg plugin has icons for `mobile`, `tablet`, and `desktop` that would be nice to use. But I can't recall if those are in 5.4. 

If they are, they'll be in packages/icons/src/library/mobile.js, tablet.js and desktop.js, and we could use those instead of the text labels.

Another option is to choose a single label that doesn't change, such as "Breakpoint". But that doesn't seem as good. 

A third option is to close the popover as soon as an option is selected, but that seems hacky as it's not the default behavior. 

Finally we can accept the wobble and create an upstream ticket and/or patch to fix it there — basically, the heuristic for the dropdown menu should be that if it is set to open from top left, its position should be calculated from the left side of the opener, not the center, as it is now. 

Which of those options would you prefer?

johngodley:
The icons seem to work fine in 5.4:

image

Ok to go with that?

5.4 doesn't like the `isAlternate` prop though:

`React does not recognize the `isAlternate` prop on a DOM element. If you intentionally want it to appear in the DOM as a custom attribute, spell it as lowercase `isalternate` instead. If you accidentally passed it from a parent component, remove it from the DOM element.
`

jasmussen:
Let's go with the icons and remove the prop, then revisit! Thanks John.

PR #80, opened 2020-04-24
Polish Event

jasmussen:
Since making the very early designs, I've doubled back on some of the concepts, so this reduces quite a bit and puts the onus on the theme:

event polish


jasmussen:
@ajlende @mkaz here are some block directory assets you can use:

banner-772x250

banner-772x250@2x

icon-256x256

icon-256x256halfsize

jasmussen:
Awesome!

PR #81, opened 2020-04-24
Allow touch events to go into the grid

### LINKED ISSUE #70: Grid Block: Multiselect for blocks inside stopped working
Observe:

multi select

If you click and drag across blocks to select multiple blocks, only a single block is selected. 

If you hold shift and use arrow keys to select, that still works.

It's possibly a conflict with onmouseup hijacking something it shouldn't. 

@johngodley any insights?

johngodley:
This enables multi-select in inner blocks

Fixes #70

### DIFF
--- a/blocks/layout-grid/src/grid.scss
+++ b/blocks/layout-grid/src/grid.scss
@@ -56,9 +56,6 @@
 	display: flex;
 	flex-direction: column;
 
-	touch-action: none;
-	user-select: none;
-
 	@for $i from 1 through 12 {
 		@for $x from 1 through 4 {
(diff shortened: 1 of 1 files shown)


johngodley:
@jasmussen this seems a simple fix, but I don't know if these rules are needed for something specific. Do you know if removing them trigger some other problem?

It should be noted that we do disable touch events when in `.wp-block-jetpack-layout-grid .wpcom-resize-grid` and when `.wpcom-resize-grid__resizing` is applied, so it may not be necessary.

I didn't notice any problems in a quick test.

Issue #13, opened 2019-12-16
Image Tools: Consider exploring more advanced SVG filters


johngodley:
(copying from another repo)

Working on the Image Filters I’ve run into a bit of a challenge.

The img element is kind of a weird one in the way it behaves, and the very best filters (google CSSgram) rely on literal overlays that leverage mix-blend-mode. But overlays run into trouble, because they have to be abs-positioned, which means the actual image becomes something else — it kind of has to be block level, or weird things happen:

https://codepen.io/joen/pen/vYYbwpL

However I found we can also use SVG filters to perform matrix filters on them:

https://codepen.io/joen/pen/dyyaELM?editors=1100

This means we don’t have to apply a wrapper to the img, we can just apply the filters directly. However for SVG filters to work we have to output actual markup in the document, for example this:

<svg viewBox="0 0 10 10" width="0" height="0" xmlns:xlink="http://www.w3.org/1999/xlink">
    <defs>
        <filter id="c1">
            <feColorMatrix in="SourceGraphic"
                                         type="matrix"
                                         values="1 1 0 0 0.1
                                                         1 0 0 0 0
                                                         0 0 0 1 0
                                                         0 0 0 1 0" />
        </filter>
    </defs>
</svg>
Even though this SVG is entirely invisible, it still needs to be output in the DOM both on the frontend and backend, or the filter won’t take.

Let's explore what's feasible here, as the filters we can make using only filter may be limiting.

johngodley:
Related to #12

johngodley:
Closing this as it has since moved into core.

Issue #12, opened 2019-12-16
Image tools: add duotone filters


johngodley:
(copying from another repo)

Add support for duotone filters, based on a suggestion by @pablohoneyhoney

https://duotone.shapefactory.co/

Codepen from @jasmussen showing this in action:

https://codepen.io/joen/pen/dyyaELM?editors=1100

johngodley:
> Just to clarify, that isn't actually duotone, but I believe we need duotone to make this work. I will make a new codepen.

johngodley:
(from @jasmussen)
Alright I created a new codepen, https://codepen.io/joen/pen/OJPbJrX?editors=1100.

This is duotone in that the image is using only two colors and shades between them.

It uses SVG filters still, which is good because that means we can apply it to any single image on the page, without having to wrap the image in a container or use CSS blend modes. It also has good support.

The tricky bit is, it uses a color matrix to make this happen, which to my eyes looks somewhat complicated:

<filter id="linear">
    <feColorMatrix
      type="matrix"
      values="R 0 0 0 0
              0 G 0 0 0
              0 0 B 0 0
              0 0 0 A 0 "/>
  </filter>
</feColorMatrix>
I would recommend this read on A List Apart for details.

In the codepen, I put together a couple of examples. Here's a SVG color matrix that makes the image red and blue duotone:


<filter id="c1">
	<feColorMatrix type="matrix" color-interpolation-filters="sRGB"
		values="1 0 0 0 0
			0 0 0 0 0
			-1 0 0 0 1
			0 0 0 1 0" />		
</filter>
Additionally, as we discussed, it requires us to output the actual SVG filter inline somewhere in the page that uses the filter on an image. Presumably this also means if you have multiple images with different filters, you'll need multiple SVGs with different matrices.

I defer to you, @johngodley, for whether this is within reach, do let me know. I think it's an incredible feature, and a powerful way to bring some dynamic colors to a website.

johngodley:
> I added a second image that shows a div with the image as a CSS background, with the same filter applied. The good news is, SVG filters work there too. The bad news is, if we want to apply it to CSS backgrounds without any child elements (such as text) not also inheriting it, we have to use backdrop-filter, which is not that widely supported.

> https://codepen.io/joen/pen/OJPbJrX?editors=1100

johngodley:
> Oh, and backdrop-filter works really badly in Safari too :(

> But still cool for Images!

johngodley:
[...]

PR #88, opened 2020-05-05
Try SVG-based duotone filter

ajlende:
@jasmussen This is what I was able to get done today. It's still very broken, but I found a way to get the server-side rendering mostly working.

The ids are a little wonky, the filter is applied to the placeholder, it bleeds outside of the image in Chrome, it doesn't render at all in Firefox, and there isn't a way to disable them at the moment. But there _are_ duotone filters and you _can_ change the color of them. 🙂

There's a lot of other effects that can be done and this technique could be applied to _any_ block. Going the SVG filter route also opens up the possibility of tweaking things like gamma and contrast on the duotone and adding more colors to the gradient. 

Related to #12 and #13

duotone

image

image


jasmussen:
OMG this is so good.

I'm going to see what tweaks I can push!

Screenshot 2020-05-05 at 09 38 04

jasmussen:
This is really cool. I'm impressed you managed to build this as a plugin. 

One challenge, though — the filter should ideally be attached to only the element that is to receive the filter. In the case of the Image block, it's the `img`. That will also avoid the resize handles and caption being filtered.

In the case of the Cover block it's a bit more tricky as it's the block itself that holds the background image, i.e. the filter will affect child blocks also. We could potentially start with the Image block and then refactor the cover block to be more amenable to this type of filtering. 

I managed to fix that extra padding around the image, that was curious indeed and to be quite honest I'm not sure why my fix works. 

I created https://codepen.io/joen/pen/RwWxZpN?editors=1100 to debug this — it appears the `feColorMatrix` was the cause of it. When I use the identical color matrix as I created for the previous test codepen, everything is fine, even the colors. Look:

extra padding

Again, I don't know why this works, but it appears to.

jasmussen:
I managed to get it working in Firefox.

Turns out, for an element to be able to be filtered by an SVG filter, that SVG must not be `display: none;`. So I applied an array of other tricks that should ensure the SVG is invisible to the user, but technically visible to the filter.

Screenshot 2020-05-05 at 10 36 09

ajlende:
I think I know why your fix worked. I thought it would be okay to set the alpha to a constant `1` by making the last row `00001` changing it back to `00010` uses the alpha from the image. When they're calculating the filter, they must be doing it on a canvas larger than the image (maybe a power of two size in pixels for efficiency when sending to the GPU). The padding shows the edge of that canvas. 

The special values I had in the rows above are weighted values for producing a more natural looking grayscale instead of just taking the average of the values. They're the values for calculating luminance when converting to a YUV color space https://www.mathworks.com/help/matlab/ref/rgb2gray.html#algorithms

jasmussen:
Too awesome, feel free to restore your weighted values! I did not mean to overwrite your work, so absolutely go with the better values if you hvae the larger matte worked out.

ajlende:
The block filter approach can only apply duotone to an entire block at a time, so I'll be opening another PR with a standalone block with only the image part filtered

PR #90, opened 2020-05-07
Add a few small fixes.

jasmussen:
Extra vertical spacing is reported in Varia themes, that use this block a lot:

https://github.com/Automattic/themes/issues/1798

This is because in the editor, every single block, including block containers, has an intrinsic top and bottom margin.

This unsets that margin for the layout grid, making sure editor and frontend look the same:

Screenshot 2020-05-07 at 14 55 59
Screenshot 2020-05-07 at 14 56 02

It has a side-effect — this doesn't look great:

Screenshot 2020-05-07 at 14 56 08

However, it's still accurate! So in that vein I'm tempted to keep that as is. We _can_ make rules to space this out, but I lean towards not doing that. If you really want to insert multiple layout grids after each other, it feels correct to show that they will snug up against each other.

Finally, it fixes an issue where on some themes, the end-gutter option wouldn't work. Pictured below, end gutters not working:

layout grid

It works in this PR, rest assured.

I would not call a release of this urgent — fine to bundle it up with other patches that may need to be made.

### DIFF
--- a/blocks/layout-grid/editor.scss
+++ b/blocks/layout-grid/editor.scss
@@ -10,14 +10,20 @@
 // Unset many of the paddings and margins the block editor sets.
 // This is to make sure resizing the column containers works as expected.
 
-// 1. Reset margins on immediate innerblocks container.
+// 1. Reset margins on block itself, so it doesn't inherit the baseline margin.
+[data-type="jetpack/layout-grid"] {
+    margin-top: 0;
+    margin-bottom: 0;
+}
+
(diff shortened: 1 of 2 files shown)


jasmussen:
Rebased and tested again. This branch:

Screenshot 2020-06-09 at 11 36 15

Master:

Screenshot 2020-06-09 at 11 36 36

So yes, this is still necessary, a good visual consistency improvement.

johngodley:
This works as described, merging.

PR #95, opened 2020-06-04
Fix layout-grid to only load CSS when needed

josephscott:
The current layout grid plugin - https://wordpress.org/plugins/layout-grid/ -
will add the following to every page view as soon as it is activated:

- layout-grid/style.css 4 KB ( 562 bytes gz )
- layout-grid/blocks/front.css 61.8 KB ( 3.4 KB gz )

This addresses this problem by moving the addition of the CSS to be part of
the `render_callback` flow in `register_block_type`.

### DIFF
--- a/blocks/layout-grid/index.php
+++ b/blocks/layout-grid/index.php
@@ -3,21 +3,29 @@
 add_action( 'init', function() {
 	register_block_type( 'jetpack/layout-grid', [
 		'editor_script' => 'block-experiments',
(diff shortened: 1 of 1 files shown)


josephscott:
That would likely be fine as well, as my goal is all about only loading the resource when it is going to be used.

mkaz:
As `init` it should be fine since it is just registering the script, and you don't have to worry about the ordering of the enqueue being called with the render_callback.

Issue #52, opened 2020-03-13
Layout Grid: Vertical alignment option


iamtakashi:
It'd be great if we allow controlling vertical alignments like the column block from core.

I wanted to make a block pattern like this:
![tiled-gallery](https://user-images.githubusercontent.com/908665/76649992-a2708800-6559-11ea-9e4f-33f86aff3e93.png)

But at the moment:
![tiled-gallery-2](https://user-images.githubusercontent.com/908665/76650112-d8157100-6559-11ea-9cd9-766a14d5a4a1.png)

Layout Grid is so much better than the one from core, and I'd love to not give up all the advantages just because of the lack of vertical alignment :)

cc @ianstewart @alaczek

iamtakashi:
In addition to top, middle, and bottom, it might be handy to have an option for an equivalent to flexbox's `space-between` for when there is more than one element in a column.

![Pattern 4](https://user-images.githubusercontent.com/908665/76863422-5eca9680-6857-11ea-8b2a-201d0493897e.png)

johngodley:
I've started exploring this in #100

>  it might be handy to have an option for an equivalent to flexbox's space-between 

Gutenberg provides the vertical alignment control, so a `space-between` would need to be added there and it would then be available to all blocks.

PR #97, opened 2020-06-08
Fix plus.

### LINKED ISSUE #91: Layout Grid: No visible + sign when adding a new block, Conflicting with Gutenberg Plugin
Originally reported on https://github.com/Automattic/wp-calypso/issues/42540 and 9F6qB-p2

It appears that there are issues adding new blocks within an existing layout grid. There's also a conflict between the Gutenberg plugin and Layout Grid regarding styling as discovered by @edequalsawesome 

> .wp-block:not(.is-selected):not(.has-child-selected) .block-editor-default-block-appender {
    display:none
}

@serabi's notes:

#### Steps to reproduce

1. Add a Layout Grid block to the page. 
2. Add a content block (such as an Paragraph or video block) at the top of the first Layout Grid column
3. Try to add a second block, below the first block. There is no plus sign. To add the second block, you need to hit enter or click on the block you've already added and select the option to add before or add after, or you need to click in the exact spot the plus sign would be, if it is visible. 

Deactivating the Gutenberg plugin seems to help at this time. 

#### Screenshot / Video

Test Site

#### Context / Source

#user-report - see 2992383-zen

Internal discussion on Atomic Requests: 9F6qB-p2

jasmussen:
Fixes #91.

### DIFF
--- a/blocks/layout-grid/editor.scss
+++ b/blocks/layout-grid/editor.scss
@@ -63,6 +63,15 @@
 	margin: 28px 0;
 }
 
+// Override and adjust rule inherited from the core project.
+.wp-block:not(.is-selected):not(.has-child-selected) .block-editor-default-block-appender {
+	display: inherit;
+}
+
+[data-type="jetpack/layout-grid"]:not(.is-selected):not(.has-child-selected) .block-editor-default-block-appender {
+	display: none;
+}
+
 
 /**
  * Inspector Controls


johngodley:
I can't see any difference here, but it's entirely possible that I don't know how to trigger the block appender.

Also, are we ok changing the core appender like this?

`.wp-block:not(.is-selected):not(.has-child-selected) .block-editor-default-block-appender {`

jasmussen:
It's a little messy. The appender plus shows at the end of every nesting container, **if the preceeding block is not a text block**, and only when the nesting container is selected.

So if you insert a group block, and a paragraph inside, you don't see the appender. If you insert a group block and an image, you see the appender after the image if you select the group. It should be the same with the layout grid block.

The problem with the layout grid block is that the wrapping div with the `wp-block` classname is an extra wrapper around our stuff. So the structure is `.wp-block > [data-type="jetpack/layout-grid"]`, where if it was `.wp-block[data-type="jetpack/layout-grid"]` we wouldn't have to change the core behavior. 

How hard is it to shift those things around?

johngodley:
> if the preceeding block is not a text block

Bingo! I've just been using paragraphs.

We don't have much control over how inner blocks are presented. However, there appears to be a new `__experimentalTagName` and `__experimentalPassedProps` property for them which may help. I'm looking at this as part of #52.

Let's merge this now and we can revisit later.

PR #98, opened 2020-06-08
Fix gridlines

### LINKED ISSUE #89: Layout Grid: Allow gridlines to be toggled
image (1)

It can be pretty easy to get to a point where the gridlines are more distracting than helpful. For people creating patterns or themes, the Grid Layout guides are likely helpful.

However, for the end-user who is adding and using these patterns, the guides are somewhat cryptic (notice they don't line up to the actual columns) and distracting. At first glance, the guides appear to be a rendering bug and its not obvious when/why they appear — this all leads to confusion when previewing the output on the site's frontend.

I'd be nice to have a way to toggle the guides on an off, and for WPcom specifically, it might be nice to be able to turn them off by default.

jasmussen:
Fixes #89. 

Previously, the gridlines were shown if you had selected the Layout Grid block, or any child.

After this PR, the gridlines are shown only if you have selected the Layout Grid block, not when a child is selected.

Before, Child:

Screenshot 2020-06-08 at 11 37 43

After, Child:

Screenshot 2020-06-08 at 11 37 52

After, parent:

Screenshot 2020-06-08 at 11 37 56

### DIFF
--- a/blocks/layout-grid/src/grid-overlay.scss
+++ b/blocks/layout-grid/src/grid-overlay.scss
@@ -12,9 +12,7 @@
 	// By being relative when unselected and absolute when selected, we
 	// constantly "refresh" this hack.
 	position: relative;
-	.is-hovered &,
-	.is-selected &,
-	.has-child-selected & {
+	.is-selected & {
 		position: absolute;
 	}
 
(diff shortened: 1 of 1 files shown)


johngodley:
Confirmed!

PR #100, opened 2020-06-10
Add vertical alignment to grid

### LINKED ISSUE #52: Layout Grid: Vertical alignment option

johngodley:
Adds a vertical alignment button to the grid container, and the same to columns:

image

As with the columns block, changing the container alignment updates all columns. Changing a column only affects that column.

Top:

image

Centre:

image

Bottom:

image

There is some difference between having top vertical alignment applied, and having no vertical alignment. With no alignment, a column fills the height:

image

With top alignment it doesn't fill height:

image

Fixes #52

### DIFF
--- a/blocks/layout-grid/editor.scss
+++ b/blocks/layout-grid/editor.scss
@@ -214,6 +214,18 @@
  * Individual column alignment for the editor
  */
 .wp-block-jetpack-layout-grid {
+	&.are-vertically-aligned-top [data-type="jetpack/layout-grid-column"].wp-block {
+		align-self: flex-start;
+	}
+
+	&.are-vertically-aligned-center [data-type="jetpack/layout-grid-column"].wp-block {
+		align-self: center;
+	}
+
+	&.are-vertically-aligned-bottom [data-type="jetpack/layout-grid-column"].wp-block {
+		align-self: flex-end;
+	}
+
 	// Note we add the alignment on the parent block as the editor nesting means it's not possible to affect the grid
 	@for $x from 1 through 4 {
 		&.column#{ $x }-grid__valign-top [data-type="jetpack/layout-grid-column"].wp-block:nth-child(#{ $x }) {
--- a/blocks/layout-grid/editor.scss
+++ b/blocks/layout-grid/editor.scss
@@ -209,3 +209,21 @@
 		}
 	}
 }
+
+/**
+ * Individual column alignment for the editor
+ */
+.wp-block-jetpack-layout-grid {
+	// Note we add the alignment on the parent block as the editor nesting means it's not possible to affect the grid
+	@for $x from 1 through 4 {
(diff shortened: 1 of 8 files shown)


johngodley:
@jasmussen I need to do a bit of tidying up with this PR, but before I go further I'd appreciate your thoughts about the difference between no alignment and top alignment.

Does the described no/top alignment behaviour make sense (from a user perspective, not CSS)?

As a user would you expect the height to always be filled, and the vertical alignment to change the text inside the paragraph? Is this even an issue? Without the paragraph background colour you wouldn't notice the difference, and you can always set the background colour on a column itself.

jasmussen:
> Does the described no/top alignment behaviour make sense (from a user perspective, not CSS)?

If I'm understanding the question correctly, it could be applied to the paragraph text alignment box also: that's the difference between no alignment and left aligned?

The answer is that no alignment is literally unset, which means outside conditions (such as the theme) can set it via CSS. Whereas when it's explicitly set, it overrides those

jasmussen:
Ack I wrote on mobile, so accidentally posted the comment half finished. I've since edited, but in case you read comments on the email notifications be sure to jump out and read the edited comment.

johngodley:
👍 thanks, that answered my question! The current behaviour seems to be correct. I'll get this cleaned up.

jasmussen:
💯

blocks/layout-grid/editor.scss
```diff
@@ -209,3 +209,33 @@
...
+
+/**
+ * Individual column alignment for the editor
+ */
+.wp-block-jetpack-layout-grid {
+	&.are-vertically-aligned-top [data-type="jetpack/layout-grid-column"].wp-block {
+		align-self: flex-start;
+	}
+
+	&.are-vertically-aligned-center [data-type="jetpack/layout-grid-column"].wp-block {
```
jasmussen, line 221: Should we rename these though? I know the grammar may not be as pure, but to me `is-vertically-aligned-*` sounds better than `are-vertically-aligned-*`, even if it refers to the content inside the column rather than the column itself. This is what the columns block also uses: `is-vertically-aligned-bottom`

johngodley: Both names are taken from the columns block - `is-vertically-aligned` is for an individual column, and `are-vertically-aligned` is for the column block itself.

https://github.com/WordPress/gutenberg/blob/master/packages/block-library/src/columns/style.scss#L77

jasmussen: Well then. Ignore me!

PR #101, opened 2020-06-10
Use Gutenberg preview device for grid block

### LINKED ISSUE #56: Layout Grid: Explore responsive dropdown integration
Gutenberg 7.7 ships with a new responsive preview feature that lets you actually preview responsive breakpoint changes.

We should explore how the layout grid block can leverage this. Can the tabs in the sidebar invoke a particular breakpoint? Or can we contextually change the contents of the sidebar depending on which breakpoint is chosen by the user?

johngodley:
Uses the `__experimentalSetPreviewDeviceType` (Gutenberg 7.7) in the grid block instead of keeping our own internal device state.

This means the grid block now adjusts to the current preview type. Here the preview mode is changed to mobile, which changes the grid block to show the mobile view, and updates the sidebar to select the mobile tab:

image

Likewise changing the tab in the sidebar, or changing the layout mode in the grid block toolbar updates the editor device type.

Fixes #56

### DIFF
--- a/blocks/layout-grid/src/grid/edit.js
+++ b/blocks/layout-grid/src/grid/edit.js
@@ -56,9 +56,6 @@ class Edit extends Component {
 		super( props );
 
 		this.overlayRef = createRef();
-		this.state = {
-			selectedDevice: getLayouts()[ 0 ].value,
-		};
 	}
 
 	/*
@@ -79,13 +76,9 @@ class Edit extends Component {
 		this.props.updateColumns( this.props.columns, columns, columnValues );
 	}
 
-	onChangeDevice = ( selectedDevice ) => {
-		this.setState( { selectedDevice } );
-	}
-
 	onResize = ( column, adjustment ) => {
 		const { attributes, columns } = this.props;
-		const grid = new LayoutGrid( attributes, this.state.selectedDevice, columns );
+		const grid = new LayoutGrid( attributes, this.props.deviceType, columns );
 		const adjustedGrid = grid.getAdjustedGrid( column, adjustment );
 
 		if ( adjustedGrid ) {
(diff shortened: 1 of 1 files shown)


jasmussen:
So I quickly confirmed it was a core issue in reproducing it outside the layout grid block. And then I opened a GIF recorder to capture it for a ticket, and now I can't reproduce. Fingers crossed it won't return, but it's not an issue here.

jasmussen:
This is not a lot of code, and it's pretty amazing:

responsive

It does show a direct correlation between what breakpoint you're editing, and a preview in context. My instinct is that this is a really good change, so good it should go in the core project. From the GIF above, I even agree that we could consider removing that toolbar button after all, because in the responsive views it moves to the top toolbar, and up there it's really duplicating the dropdown. We also added it originally because we did not have any correlation between breakpoint in the sidebar, and what you saw.

This is a big change, I'd love a sanity check by @iamtakashi who I know has used this feature a fair bit: do you feel this feature is as helpful as I think it is?

If yes, then we should try it out, and if it goes well, consider donating it to the core project. I imagine many small block developers have wanted something similar to this.

johngodley:
> consider donating it to the core project

I don't know if there's anything here to actually donate, other than as an example. The code is already in core, and this just plugs it into the block.

jasmussen:
> I don't know if there's anything here to actually donate, other than as an example. The code is already in core, and this just plugs it into the block.

I believe the core project has a DimensionControl, and while I'm not a big fan of it, and yearning ideally for something more directly fluent, making sure that core control taps into it might be all it takes.

iamtakashi:
This is a lovely change. I find it convenient when I specify responsive patterns for the grid. Nice work!

jasmussen:
I'm actually seeing some alignment issues with handles on the other breakpoints, maybe hold off on merging for a bit as I look into that:

Screenshot 2020-06-12 at 10 19 58

jasmussen:
False alarm. This was horrendous debug CSS from my own work in progress theme that broke things. When I remove it, all is well. Continue shipping, John, and nice work:

Screenshot 2020-06-12 at 11 35 44

Screenshot 2020-06-12 at 11 35 49

Issue #63, opened 2020-03-27
Layout Grid: Allow overlapping of layers


melchoyce:
Would be cool if we could use this block to create something like this design:

![Screenshot_2020-03-26 Madies Lyophilized Fruits Que Gusto](https://user-images.githubusercontent.com/2846578/77773410-f2eae980-701f-11ea-89a2-047791c132e8.jpg)

johngodley:
I'm not sure if this kind of design is possible without a big rewrite of the block. @jasmussen do you have any thoughts?

jasmussen:
When I first saw this ticket, I thought yes — this is a cool benefit and an obvious improvement. I imagined that it was a matter of simply adjusting the column spans and the overlap would happen naturally.

And... that does seem to be _possible_ actually:

![overlap](https://user-images.githubusercontent.com/1204802/80216033-b7066c80-863d-11ea-8bb9-f68e2112df53.gif)

It _appears_ to be even simpler than that GIF, if I apply `position: relative; z-index: 1;` on one of the columns, it naturally floats above adjacent columns without collapsing the parent container (which `position: absolute;` does). So it seems like it might not need a total rewrite — for starters we could experiment with a toggle that simply elevates the z-index of a single column, and see how that fares in the branch.

I'm sure it's not quite as simple, UI wise — I believe John that you've built in some resizing smarts into the UI that would probably need to be toggled off for the column that's chosen to "float". 

Implicitly this is also still the inherent limitation that we are only using columns at the time, so you couldn't actually create what's shown in Mel's mockup with an item that's vertically centered above text... or well maybe. But it's not quite as freeform. 

But it seems worth trying as a small prototype in a PR and then decide the next step. What do you think John?

johngodley:
I've been working on something unrelated and I think it might end up being useful here. The unrelated change moves from using grid offsets to grid starts (i.e move from relative to absolute grid positioning). For the end user this shouldn't have any effect, but it would make overlapping layers easier as two columns could have the same absolute grid position.

With this in mind, it seems worth thinking a bit more about what is involved.

From the above experiment it seems it's technically possible to overlap a grid column using CSS. We should do some cross-browser tests to confirm this is expected behaviour and not some side-effect.

In terms of UI we could allow a column to be 'unlocked'. This would remove it from the existing grid validation, and allow the user to change the start value to anything. If we detect an overlap then we apply the grid float.

I imagine we could combine this with drag-and-drop so the column can be dragged around the grid.

[...]

PR #105, opened 2020-06-25
Grid block refactor

johngodley:
This is a big refactor that does several things:

- It moves from `span` and `offset` to `span` and `start` - columns are absolutely positioned instead of being relative to the previous column. The user continues to see things in terms of `offset`
- Columns now contain their own `span` and `start` attributes instead of storing everything in the grid container

This results in many benefits:
- Big reduction in complexity. We no longer have to convert between `start` and `offset` all the time
- CSS size is greatly reduced as we can apply the same CSS class to each column, instead of having to provide every possible combination on the parent
- Possibility of using absolute positioning for future things (see #63)

Other changes:
- Column resizing is now squishy. That is, it will squash columns if you resize into them. This contrasts to the previous version where it would use spare-space, if available, but wouldn't squash.
- It's no longer possible to move a column onto a different row. Previously, using the inspector only, you could modify the offset and span values by entering specific values that meant a column could be pushed onto another row. I don't think anyone knew this was possible, and it wasn't very easy to use. #63 and drag/drop columns could be a better way of changing columns, so I don't think the loss here is a problem

Upcoming changes:
- Unit tests to check the grid continues to behave as expected
- Block migration. The block attributes have changed, and we need to make the block not cause invalidations. Ideally it can transparently accept old attributes and convert them to new ones when re-saved.
- Further CSS size reduction. Maybe we can split out the IE prefix CSS into a separate IE only file?
- Merge `front.css` and `style.css` together. I know we split them apart, but it makes integration difficult, and it would be better if we merge them
- Changing the layout auto-selects the first column. Not sure why, and it's annoying


johngodley:
@jasmussen this is still a work in progress and doesn't immediately need any attention, but just putting it on your radar.

jasmussen:
This is a big one. I've received your ping, and it's on my ToDo, and I'll absolutely get to it. But I want to do it right.

jasmussen:
I tested this today, and I must say, the amount of red code here is impressive.

The block itself is also working really well. In fact, it seems to be working perfectly, except for a few issues that I believe are not related to this branch specifically, but rather the block itself. But perhaps simply due to testing this heavily, they appeared.

The bad part is, I can no longer reproduce the issues, either in master or this branch. But nevertheless, I did encounter an issue where I switched from 3 to 4 columns, then pressed Undo, and the layout was messed up rather badly. Behold these values:

Screenshot 2020-07-10 at 12 05 57

I don't know if I didn't properly clear cache, but I could not recreate the conditions to cause this. But it did involve undo/redo and switching the amount of columns. 

But in every case, even when the column values got broken for inexplicable reasons, the CSS seemed completely correct, insofar as the frontend always matched the backend.

What is the next step to moving forward with this one? Do we need some Takashi testing? Did you need CSS input?

PR #108, opened 2020-06-29
Fix loading of style.css inside the editor

### LINKED ISSUE #103: Layout Grid: Missing Outermost Gutters in Editor
The outermost gutters seem gone now in the Editor, and the content inside bumps the edges of the browser screen. I don't think it's been like this for a long time...

Screenshot 2020-06-18 at 12 17 13

The gutters appear fine on the front of the site.

Screenshot 2020-06-18 at 12 17 20

johngodley:
The previous change to `render_callback` meant that `style.css` didn’t load inside the editor (no blocks are rendered).

Go back to using the normal `style` param when registering the block, and let Gutenberg decide when to load the style

Fixes #103

### DIFF
--- a/blocks/layout-grid/index.php
+++ b/blocks/layout-grid/index.php
@@ -4,9 +4,9 @@
 add_action( 'init', function() {
 	register_block_type( 'jetpack/layout-grid', [
 		'editor_script' => 'block-experiments',
+		'style' => 'block-experiments',
 		'editor_style' => 'block-experiments-editor',
 		'render_callback' => function( $attribs, $content ) {
-			wp_enqueue_style( 'block-experiments' );
 			wp_enqueue_style( 'wpcom-layout-grid-front' );
 			return $content;
 		},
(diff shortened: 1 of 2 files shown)


simison:
Does the style still load on frontend only when the block is rendered? :-)

Alternatively you could `@import` frontend style to the editor style file.

simison:
FYI @Automattic/ganon would be good to check if template API still needs D45267-code

johngodley:
> Does the style still load on frontend only when the block is rendered

This change makes it load `style.css` the standard Gutenberg style loading route.

> would be good to check if template API still needs D45267-code

It won't affect that, which refers to `front.css`, which is only loaded when the block is rendered.

simison:
> > Does the style still load on frontend only when the block is rendered
>
>This change makes it load style.css the standard Gutenberg style loading route.

Right, that's been problematic for performance on .com and reason why it was changed in first place here. :-) We don't have this problem on Jetpack for example because we use combination of sass imports and `render_callback`

johngodley:
> problematic for performance on .com and reason why it was changed in first place here

It was specifically `front.css` that is a problem for the grid block as this is a non-standard file, and is of a significant size. 

> Imagine if every WordPress page would load CSS+JS for every block ever shipped on .com

Sure, I get that, and it affects every site not just those on WordPress.com. Gutenberg is working on this (https://github.com/WordPress/gutenberg/pull/22754) so the situation will hopefully improve.

With this change we are in a better performing position than the block has been in from the start (i.e. `front.css` is still optionally loaded), and the editor is now fixed (the reason for making this change).

It may not be the most performant position, but this will be improved in future work.

simison:
Thanks for elaborating! :-)

andrewserong:
> would be good to check if template API still needs D45267-code

Just to be sure, I tried testing the endpoint with that diff reverted, and it looks like we still need it.

PR #110, opened 2020-07-04
Update blocks categories

fullofcaffeine:
#### Changes proposed in this Pull Request:

- Update the category names for relevant blocks in `block-experiments`. 
-  Adds a helper function that is used to safely update category names while still supporting old environments

#### Related PRs
* Jetpack blocks: https://github.com/Automattic/jetpack/pull/16363
* FSE blocks (wp-calypso): https://github.com/Automattic/wp-calypso/pull/43670
* wpcom-blocks: https://github.com/Automattic/wpcom-blocks/pull/178

##### Atomated tests

* `yarn test`

##### To manually test the changes

1. Setup a local Wordpress instance (i.e by using `wp-env`)
2. Follow the instructions to build `block-experiments` in its README
3. Symlink or move it to the plugins folder of your WP instance
4. Activate the plugin
5. Profit

#### Categories

* [X] Keep `a8c/bauhaus-centenary` in `widgets`
- [X] Keep `a8c/event` in `widgets`
- [X] [Change](https://github.com/Automattic/block-experiments/pull/110/files#diff-264f12914f7ca9ea517358bfc96cedaeR51) `jetpack/layout-grid` from `layout` -> `design`
- [X] [Change](https://github.com/Automattic/block-experiments/pull/110/files#diff-264f12914f7ca9ea517358bfc96cedaeR120) `jetpack/layout-grid-column` from `layout` -> `design`
- [X] Keep `a8c/motion-background-container` in `widgets`
- [X] Keep `a8c/starscape` in `widgets`
- [X] Keep `a8c/waves` in `widgets`

~Note: If [this PR](https://github.com/WordPress/gutenberg/pull/23695) gets approved and merged, then I'll remove the helper function and tests from here (including the installation/configuration of jest).~

### DIFF
--- a/package.json
+++ b/package.json
@@ -18,7 +18,8 @@
     "clean": "rm -rf ./build ./plugin",
     "lint": "npm run lint-js & npm run lint-php",
     "lint-js": "wp-scripts lint-js blocks",
-    "lint-php": "composer run lint"
+    "lint-php": "composer run lint",
+    "test": "wp-scripts test-unit-js"
   },
   "devDependencies": {
     "@babel/plugin-proposal-class-properties": "^7.10.1",
--- a/blocks/_helpers.js
+++ b/blocks/_helpers.js
@@ -1,0 +1,39 @@
+/**
+ * WordPress dependencies
+ */
+import { getCategories } from '@wordpress/blocks';
+
+/**
+ * Accepts an array of category names and returns the first one that
+ * exists in the categories returned by `getCategories`. This allows
+ * for a "graceful degradation" strategy to category names, where
+ * we just add the new category name as the first item in the array
+ * argument, and leave the old ones for environments where they still
+ * exist and are used.
+ *
+ * @example
+ * // Prefer passing the new category first in the array, followed by
+ * // older fallback categories. Considering the 'new' category is
+ * // registered:
+ * getCategoryWithFallbacks('new','old', 'older');
+ * => 'new'
+ *
+ * @param {string[]} requestedCategories - an array of categories.
+ * @return {string} the first category name found.
+ * @throws {Error} if the no categories could be found.
+ */
+export function getCategoryWithFallbacks( ...requestedCategories ) {
+	const knownCategories = getCategories();
+	for ( const requestedCategory of requestedCategories ) {
+		if (
+			knownCategories.some( ( { slug } ) => slug === requestedCategory )
+		) {
+			return requestedCategory;
+		}
+	}
+	throw new Error(
+		`Could not find a category from the provided list: ${ requestedCategories.join(
+			','
+		) }`
+	);
+}
--- a/blocks/layout-grid/src/index.js
+++ b/blocks/layout-grid/src/index.js
@@ -14,15 +14,26 @@ import saveGrid from './grid/save';
 import editColumn from './grid-column/edit';
 import saveColumn from './grid-column/save';
 import { GridIcon, GridColumnIcon } from './icons';
-import { getSpanForDevice, getOffsetForDevice, DEVICE_BREAKPOINTS, MAX_COLUMNS } from './constants';
+import {
+	getSpanForDevice,
+	getOffsetForDevice,
+	DEVICE_BREAKPOINTS,
+	MAX_COLUMNS,
+} from './constants';
+import { getCategoryWithFallbacks } from '../../_helpers';
 
 function getColumnAttributes( total, breakpoints ) {
 	const attributes = {};
 
 	for ( let index = 0; index < total; index++ ) {
 		breakpoints.map( ( breakpoint ) => {
-			attributes[ getSpanForDevice( index, breakpoint ) ] = { type: 'number' };
-			attributes[ getOffsetForDevice( index, breakpoint ) ] = { type: 'number', default: 0 };
+			attributes[ getSpanForDevice( index, breakpoint ) ] = {
+				type: 'number',
+			};
+			attributes[ getOffsetForDevice( index, breakpoint ) ] = {
+				type: 'number',
+				default: 0,
+			};
 		} );
 	}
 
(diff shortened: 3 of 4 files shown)


jest.config.js
```diff
@@ -0,0 +1,3 @@
+module.exports = {
+	preset: '@automattic/calypso-build',
```
sirreal, line 2: I don't see this dependency listed. Is this working as expected? Did this config need to be included?

package.json
```diff
@@ -34,6 +35,7 @@
     "eslint-plugin-react": "^7.20.0",
     "eslint-plugin-wpcalypso": "^4.1.0",
     "fs-extra": "^9.0.1",
+    "jest": "26.1.0",
```
sirreal, line 38: We probable don't need to touch dependencies. This package appears to rely on `@wordpress/scripts`, we might as well use that tooling for testing:

https://github.com/WordPress/gutenberg/blob/master/packages/scripts/README.md

package.json
```diff
@@ -18,7 +18,8 @@
     "clean": "rm -rf ./build ./plugin",
     "lint": "npm run lint-js & npm run lint-php",
     "lint-js": "wp-scripts lint-js blocks",
-    "lint-php": "composer run lint"
+    "lint-php": "composer run lint",
+    "test": "jest"
```
sirreal, line 22: This can probably become the following and eliminate the need for dealing with `jest` ourselves.

```suggestion
    "test": "wp-scripts test-unit-js"
```

fullofcaffeine: Thanks for the suggestion, I'll make the change :+1:

package.json
fullofcaffeine: :+1:

jest.config.js
fullofcaffeine: Thanks for pointing that out!

The test task is running fine and there are no warning or errors, but that actually might have been an oversight from my part and it's probably using the vanilla jest configuration... the actual test is quite simple so it doesn't really need anything from this preset, although I think this is the preset we should use to align with the Calypso standards, AFAIK. 

I'll double-check this now and add the dependency if needed.

sirreal: The testing via wp-scripts shouldn't need any dependency changes. It would be good to remove this file, the lock changes, and the dependencies changes.

We shouldn't need to add any dependencies here.

fullofcaffeine: Copy that, thanks for letting me know! I'll clean this up and just use `wp-scripts`. I'll do the same for `wpcom-blocks` since it has a similar setup.

package.json
fullofcaffeine: Nice to know about the `wp-scripts` tool btw :+1:

blocks/_helpers/index.js
```diff
@@ -0,0 +1 @@
+export * from './helpers';
```
sirreal, line 1: We could probably do with a bit less indirection here 🙂 

Effectively, we have `_helpers` proxied to `_helpers/helpers.js` by way of `_helpers/index.js`. I appreciate re-exporting small, related modules from an index file, but this seems like [YAGNI](https://en.wikipedia.org/wiki/You_aren%27t_gonna_need_it).

I'd propose these alternatives we put the contents of `blocks/_helpers/helpers.js` directly in `blocks/_helpers.js` and remove directory (YAGNI, this is simplest).

johngodley:
> supporting old environments

I'm not sure of the context to this change, but what old environment is this referring to?

sirreal:
> I'm not sure of the context to this change, but what old environment is this referring to?

@johngodley Gutenberg 8.4 removes some block categories and introduces some new ones. There is some built-in handling that maps legacy categories to new ones (blocks using older categories won't break), there is no protection in the other direction (blocks using new categories either without Gutenberg installed or with an out-of-date version _will break_).

The purpose of this change is to define the category for each block in both contexts, whether new categories or old categories are present.

johngodley:
Thanks. If it makes things easier we can tie the release of this to minimum WP 5.5?

sirreal:
I'm not familiar with this project, but I don't think it's necessary to wait. There's a relatively straightforward helper function added.

I think this can move ahead without restricting the version to WP 5.5 unless you disagree.

johngodley:
The plugin gets bundled up for .org and for .com, and this change will require modifying that bundling process to also include the helper file (sadly it's in a different location).

This isn't a problem as such. However we don't really need a lot of backwards compatibility, and we generally set the minimum to the latest version of WP. We'll do this for some upcoming changes anyway.

Based on what you said about WP 5.5 we could simplify the change to a straight `layout` => `design` swap, drop the helpers, and bump the minimum. Once 5.5 is on .com and Atomic we can merge and use the new category.

Happy to go either way though.

sirreal:
What's the timeline like for a WP 5.5–only release? I'd be happy to drop the legacy category handling and just update to the new categories. Would you take over merging this when appropriate for that release?

blocks/_helpers/index.js
fullofcaffeine: Yeah, good points :+1:

johngodley:
WP 5.5 seems scheduled for 11th August. In terms of the layout grid we'll be releasing a minor fix in the next day or two (1.2.4), and a more substantial release in the next week or two (1.3). Both of these should make it out before WP 5.5.

There's a big release coming up (1.4), which will be ready after 5.5 is released, and will definitely require 5.5 as a minimum.

In the meantime, assuming that the existing category continues to work with 5.5, then I can include this PR as part of the bigger 1.4 release.

If something does happen in the meantime then we can just include this PR anyway.

sirreal:
I'm happy to let you decide when the tradeoffs to land this sooner or later make the most sense for this project.

fullofcaffeine:
Should we abandon this PR?

johngodley:
Yeah, sorry, Layout Grid 1.3 hasn't been released yet so this was left untouched.

Based on the above conversation, and that 5.5 is now released, I don't think we need the back-compat helpers anymore and as such I'll close this PR.

I'll ensure that the category changes is applied to all the blocks in this repo, and they'll be included in the next releases. I'll also bump the minimum supported WP to 5.5.

PR #112, opened 2020-07-08
Layout Grid: Increase specificity of the column margins and paddings reset

Copons:
When themes assign custom margins to blocks in the editor style, they eventually gain an increased specificity due to the automatically-added `.editor-styles-wrapper` class.

In some cases, even if the theme was careful enough to be as "unspecific" as possible, it can happen that its rules override some essential rules of the block.

For example, the [Seedlet theme](https://github.com/Automattic/themes/blob/55051ac7b0904213034713c8d8c56ba20a5766ec/seedlet/assets/sass/blocks/utilities/_editor.scss#L158-L174) customizes the vertical margins of all blocks, and then removes the top margin of the first inner block to avoid double margins.
Once parsed by Gutenberg, that rule manages to override the Layout Grid Column `margin: 0`, causing a misalignment where only the first column is positioned correctly:

Screenshot 2020-07-03 at 01 12 23

This PR simply increases the columns' margins and paddings reset specificity to override the theme customization.
The new specificity is not "extreme", so that if the theme _really_ needs it, it can still easily override it.

The end result would be:

Screenshot 2020-07-07 at 15 41 14

Fixes https://github.com/Automattic/wp-calypso/issues/43882

### DIFF
--- a/blocks/layout-grid/editor.scss
+++ b/blocks/layout-grid/editor.scss
@@ -23,7 +23,7 @@
 }
 
 // 3. Unset margins and paddings for column container.
-[data-type="jetpack/layout-grid-column"].wp-block { // Selector needs specificity to override.
+.wp-block-jetpack-layout-grid [data-type="jetpack/layout-grid-column"].wp-block { // Selector needs specificity to override.
 	margin: 0;
 	padding-left: 0;
 	padding-right: 0;


jasmussen:
This seems good to me. 👍 👍

Copons:
Thanks for the review and the merge! 🙇‍♂️

PR #116, opened 2020-07-15
Adjust Starscape Heading

ajlende:
Adjusts the line height for the Starscape Block. The line height is now fixed at 1.3.

@pablohoneyhoney suggested making it bold by default, but there isn't any mechanism in the RichText component to provide default formats (https://github.com/WordPress/gutenberg/issues/23976). I could force the font-weight to be bold in the CSS, but that would override any RichText bold styles applied.

Before:

Starscape before changes

After:

Starscape after changes

### DIFF
--- a/blocks/starscape/style.scss
+++ b/blocks/starscape/style.scss
@@ -21,6 +21,7 @@
 		z-index: 1;
 		width: 100%;
 		font-size: 84px;
+		line-height: 1.3;
 	}
 }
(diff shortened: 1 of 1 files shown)


pablohoneyhoney:
Looking good

oskosk:
This changeset just piggybacked on Starscape 1.0.3 released today 🥳

PR #120, opened 2020-07-23
Pass the correct script handle for translations

dd32:
Fix translations by passing the script handle to wp_set_script_translations() rather than the block name.

In testing the Block Directory in non-english, I ran into these blocks not loading translations. Turned out it's passing the block slug rather than the script to localise.

### DIFF
--- a/blocks/bauhaus-centenary/index.php
+++ b/blocks/bauhaus-centenary/index.php
@@ -16,5 +16,5 @@ add_action( 'init', function() {
 		]
 	);
 
-	wp_set_script_translations( 'a8c/bauhaus-centenary', 'bauhaus-centenary' );
+	wp_set_script_translations( 'block-experiments', 'bauhaus-centenary' );
 } );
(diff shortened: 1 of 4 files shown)


johngodley:
The blocks are designed to be released individually, not as a 'block-experiments' bundle, and they must be [packaged](https://github.com/Automattic/block-experiments#packaging-a-plugin) with the included script. This will ensure that the current `wp_set_script_translations` value is correct.

dd32:
You've misunderstood the PR. 

The build process is not changing the first parameter to `wp_set_script_translations()`, which is intended on being the Script Handle, but is currently being passed a block slug.

This PR changes that parameter to the `block-experiments` literal that the build process seems to use to set the correct handle.

johngodley:
Alright, the reason for my confusion is that I had already added the correct translation call:

https://github.com/Automattic/block-experiments/blob/master/bundler/template/index.php#L44)

But it was added somewhere else, and apparently without realising that there was an existing (incorrect) translation call added in #39. My change was unreleased so doesn't yet appear on wp.org.

I'll remove the one I added and go with the one here.

PR #124, opened 2020-07-24
Add initial blockbook

youknowriad:
If you missed my announcements, I released a tool to showcase and develop blocks in isolation https://riad.blog/2020/07/22/introducing-blockbook-for-wordpress/

Capture d’écran 2020-07-24 à 1 25 00 PM

This PR adds a blockbook in this repository:

 - Some blocks in this repository rely on experimental JS features (class properties) not supported by blockbook (relies on a setup similar to wp-scripts) which means they can't be include in BlockBook yet.
 - I didn't register any custom stories for the blocks, but I think you should consider adding different stories for the different blocks (See the "stories" tab here as an example https://youknowriad.github.io/blockbook/block/core---buttons )
 - I added a Github action to deploy it as a Github Pages for this repository.

There are still a few things that can be improved in BlockBook but I'd love to see this considered here to get more feedback on it :) 

Thanks.

### DIFF
--- a/.blockbook/index.js
+++ b/.blockbook/index.js
@@ -1,0 +1,29 @@
+
+import { registerBlockType, registerTheme } from 'blockbook-api';
+
+// Register blocks
+import * as bauhausCentenaryBlock from '../blocks/bauhaus-centenary/src';
+import * as eventBlock from '../blocks/event/src';
+import * as motionBackgroundBlock from '../blocks/motion-background/src';
+import * as starscapeBlock from '../blocks/starscape/src';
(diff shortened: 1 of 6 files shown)


jasmussen:
This is really cool and there's no reason we don't get this in as soon as we can. @mkaz or @georgeh do you have time to look at this?

georgeh:
I like it and it (mostly) works on my machine. The Waves block is erroring out for me, not sure if that's related to the other errors you mentioned.

I don't see a way to test the GitHub actions beyond merging it, but I say go for it.

PR #134, opened 2020-08-20
Create a Book block

vcanales:
**This PR includes changes made to the the dependencies on another PR: #132 and it should land after.** I considered both changes to be sufficiently different to require different discussions.

Add a Book block, with a caption that is specific to Book data, and a special placeholder for the cover image. 
This is a building block for these future additions:

- A 3D effect for the book cover, which could allow us to flip the book, or even show a more book like version of the image, like this example
- A book shelf block that will be used to group book blocks via InnerBlocks. It should display book metadata stored on the book blocks.

#### Preview
Book1
Book2


jasmussen:
Love it! Unless the other folks requested want to go with a code review more than I've done, I say let's get this in!

PR #137, opened 2020-09-02
Add Duotone Block

ajlende:


blocks/duotone/style.scss
```diff
@@ -0,0 +1,9 @@
+/* Front end styles */
+
+@import './src/_variables.scss';
```
jasmussen, line 3: Do we need to importa ll of these variables for breakpoints and mixins? I think they're already out of date with the core project.

ajlende: What I'm doing now is a sort of brute-force approach to adding duotone to the existing blocks. Not all variables in those files are used, but copying whole files makes my plans for updating easier. 

I was going to produce a patch that can be applied to files copied from Gutenberg. Then, when things update in Gutenberg, we can copy the whole files down and apply the patch.

It's not a great solution, but it seems like it could probably work.

ajlende:
Closing in favor of #143

PR #139, opened 2020-09-04
Add/model layout grid picker

enejb:
Follow up on the https://github.com/Automattic/block-experiments/pull/136

Updates the block inserted so that we show a modal that lets the user pick the number of columns that they want to see in the layout grid.

iOS demo app
Screen Shot 2020-09-04 at 11 35 49 AM

Android demo app
Screen Shot 2020-09-04 at 12 08 14 PM

**To test:** 
Follow the steps in https://github.com/wordpress-mobile/gutenberg-mobile/pull/2582

Then navigate to block-experiments directory
`cd block-experiments` and switch the repository to this branch. 

Reload the Demo app.

PR #140, opened 2020-09-04
Mobile: Add gutter settings

enejb:
Adds gutter settings to the mobile view inspector controls. 

This PR continuous from https://github.com/Automattic/block-experiments/pull/139 and add the gutter settings. 

- It currently doesn't change the gutters on block UI level but just adds the settings. 

See:
_iOS:_
Screen Shot 2020-09-04 at 1 59 15 PM
Screen Shot 2020-09-04 at 1 59 19 PM

_Android:_
Screen Shot 2020-09-04 at 2 01 32 PM
Screen Shot 2020-09-04 at 2 01 41 PM

**To test:**
Follow the steps in wordpress-mobile/gutenberg-mobile#2582

Then navigate to block-experiments directory
cd block-experiments and switch the repository to this branch.

Reload the Demo app.

### DIFF
--- a/blocks/layout-grid/src/grid/edit.native.js
+++ b/blocks/layout-grid/src/grid/edit.native.js
@@ -7,13 +7,26 @@ import { delay, flatMap } from 'lodash';
 /**
  * WordPress dependencies
  */
-import { InnerBlocks, BlockVariationPicker } from '@wordpress/block-editor';
-
+import { 
+	InnerBlocks,
+	InspectorControls,
+	BlockVariationPicker,
+ } from '@wordpress/block-editor';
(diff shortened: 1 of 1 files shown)


iamthomasbishop:
@enejb Taking a look at the screenshots above, I have a few bits of feedback:

* We can remove the "Cancel" action on the Android bottomSheet
* Can we left-align the footer text (`Note: ...`) and remove the separator above it?

I think in the long term, I'd like to consider using sub-sheets instead of ActionSheet for cells that act as select boxes, but we're currently using ActionSheets in most places so let's not stress about it as part of this work.

Issue #141, opened 2020-09-07
Layout Grid: Add `has-background` if a custom column background color is set


jasmussen:
Currently if you select a single column, and apply a preset swatch color, it receives a `has-background` class, in addition to `has-[color]-background`:

```
<div class="wp-block-jetpack-layout-grid-column wp-block-jetpack-layout-grid__padding-none has-background has-pretty-blue-background-color">
```

However is you simply set any random background color that is not a predefined swatch, that `has-background` is missing:

```
<div class="wp-block-jetpack-layout-grid-column wp-block-jetpack-layout-grid__padding-none" style="background-color: #c0a417;">
```

This causes _part_ of the issue seen in #131. 

@johngodley is this easy or hard to fix?

PR #142, opened 2020-09-07
Fix issue with layout inconsistency.

### LINKED ISSUE #131: Layout Grid: Inconsistent margins & paddings
I have noticed that the 3 column grid layout will create inconsistent margins for the middle block. This seems to be happening on Varia child themes (See Alves and Barnsbury, below):
Screen Shot 2020-08-14 at 14 12 21
Screen Shot 2020-08-14 at 18 32 54

This is also happening with TwentyTwenty, with the additional issue of uneven padding:
Screen Shot 2020-08-14 at 18 32 41

I initially noticed this on a self-hosted site but I was able to reproduce it on a simple site (Gutenberg 8.7.1)

Context 3224728-zen

jasmussen:
Partially fixes #131. (#141 fixes the rest)

#113 added a max-width to columns. This broke columns with background colors, because they work by using negative margins to make the columns wider than they appear. (i.e. a column with 24px of negative margins needs to be 100% + 48px wide in order to not show gaps)

This PR fixes that issue, by customizing that max-width per padding. It's not super clean, but it should address the situation where child contents are wider than the column.

To test:

- Try content with columns that have background colors set (use preset colors until #141 is fixed) and try with various column paddings.
- Bonus: test the above in Firefox, but with a Jepack Slideshow block inside one of the columns.

### DIFF
--- a/blocks/layout-grid/style.scss
+++ b/blocks/layout-grid/style.scss
@@ -113,6 +113,9 @@
 
 		&.has-background {
 			padding: $padding-none ($padding-none + $grid-gutter__background-offset);
+
+			// Needs a max-width in order to constrain contents.
+			max-width: calc( 100% + #{ ($padding-none + $grid-gutter__background-offset) * 2 } );
 		}
 	}
 
@@ -121,6 +124,9 @@
 
 		&.has-background {
 			padding: $padding-small ($padding-small + $grid-gutter__background-offset);
+
+			// Needs a max-width in order to constrain contents.
+			max-width: calc( 100% + #{ ($padding-small + $grid-gutter__background-offset) * 2 } );
 		}
 	}
 
@@ -129,6 +135,9 @@
 
 		&.has-background {
 			padding: $padding-medium ($padding-medium + $grid-gutter__background-offset);
+
+			// Needs a max-width in order to constrain contents.
+			max-width: calc( 100% + #{ ($padding-medium + $grid-gutter__background-offset) * 2 } );
 		}
 	}
 
@@ -137,6 +146,9 @@
 
 		&.has-background {
 			padding: $padding-large ($padding-large + $grid-gutter__background-offset);
+
+			// Needs a max-width in order to constrain contents.
+			max-width: calc( 100% + #{ ($padding-large + $grid-gutter__background-offset) * 2 } );
 		}
 	}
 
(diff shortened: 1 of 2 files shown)


aaronrobertshaw (approved):
This tests well for me 👍 

* Tested initially using the code snippets provided in https://github.com/Automattic/block-experiments/issues/131#issuecomment-682889749. 
* Also tried a range of different grid and gutter settings, everything appeared to display correctly
* Tested in Chrome, Safari and Firefox using Slideshow as inner block, all worked.

One minor issue I noticed while testing this was that columns with background colors did not "stretch" to the full column height in the editor in Chrome or Safari. This was due to the additional wrapper elements required by the block editor. As a quick fix, adding `height: 100%;` to those introduced wrapper elements allows the actual grid columns to calculate their height in Chrome and Safari.

In grid.scss I added the following and it worked ok.
```
.wp-block-jetpack-layout-grid > .block-editor-inner-blocks > .block-editor-block-list__layout {
	> .wp-block {
		height: 100%;
	}
}
```

| Before | After |
|--------|------|
| Screen Shot 2020-09-08 at 11 36 10 am | Screen Shot 2020-09-08 at 11 35 50 am |


If you feel this is better addressed via a separate PR, that's ok with me. I won't let it hold up this one.

jasmussen:
Thanks so much for thorough testing!

> One minor issue I noticed while testing this was that columns with background colors did not "stretch" to the full column height in the editor in Chrome or Safari. This was due to the additional wrapper elements required by the block editor. As a quick fix, adding height: 100%; to those introduced wrapper elements allows the actual grid columns to calculate their height in Chrome and Safari.

I can't actually reproduce that. Do you have additional steps to explain what's up here? Which version of WordPress did you test with? Which Gutenberg? I tested with the latest of everything and I'm seeing it work as is:

Screenshot 2020-09-08 at 10 03 26

Screenshot 2020-09-08 at 10 04 06

Screenshot 2020-09-08 at 10 04 36

I'm going to merge this one, but please do follow up with any steps you have and I'll make another PR to include your code snippet at that time. Thank you!

aaronrobertshaw:
@jasmussen Sorry to waste your time on that one. My WP version was still on 5.4 and Gutenberg was on branch I'd been testing previously.

After switching back to master and the latest versions, the issue I was seeing is not present. I do not believe any further action is needed. Thanks 🙂

jasmussen:
Not a single moment was wasted, on the contrary you were very helpful. And thank you for checking!

PR #144, opened 2020-09-09
Fix: Add props passing into icons

enejb:
Currently the icons don't allow for props to be passed to the main SVG component. 

The problem is that this prevents the mobile Gutenberg editor from passing in the correct fill prop for the icon.

On the mobile inserted.
Before:
Screen Shot 2020-09-08 at 12 34 53 PM

After:
Screen Shot 2020-09-09 at 12 50 15 PM

**To tests**

I build a new version of the layout-grid plugin. 
using `yarn plugin layout-grid` 
and then `yarn bundle` 

And then I installed the plugin on my site. 
And tried to find if the icons have changed anywhere with undesirable outcomes. 

See:
Screen Shot 2020-09-09 at 1 44 38 PM

Screen Shot 2020-09-09 at 1 44 11 PM

Screen Shot 2020-09-09 at 1 44 17 PM

Screen Shot 2020-09-09 at 1 44 30 PM

To test it on mobile I followed the instructions in https://github.com/wordpress-mobile/gutenberg-mobile/pull/2582 
and then patched the icon file. Which fixed how the icons appeared.

### DIFF
--- a/blocks/layout-grid/src/icons.js
+++ b/blocks/layout-grid/src/icons.js
@@ -4,7 +4,7 @@
 
 import { Path, SVG } from '@wordpress/components';
 
-export const GridIcon = ( props = null ) => (
+export const GridIcon = ( props ) => (
 	<SVG xmlns="http://www.w3.org/2000/svg"
 		width="24" height="24"
 		viewBox="0 0 24 24"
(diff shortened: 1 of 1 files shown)


mkaz (changes requested):
This looks fine, one question on the two GridIcon components you include a `props = null` but on the rest you just add a `props` parameter without the default null.

In my initial testing it looks like it doesn't make a difference, I would probably prefer without the default but either way should be consistent across all.

mkaz (approved):
Thanks for updating, looks good :+1:

enejb:
Thanks for the review and testing @mkaz !

Issue #154, opened 2020-11-10
Layout Grid: Appears broken in the full-site editor


kjellr:
Testing with Gutenberg 9.3.0, the Layout Grid block appears broken in the full-site editor. Instead of stretching to fill the container, each column only takes up a fraction of the available horizontal space. 

Post/Page Editor (As expected)|Site Editor (columns don't fill available space)
---|---
![Screen Shot 2020-11-10 at 11 16 30 AM](https://user-images.githubusercontent.com/1202812/98700493-3c394080-2346-11eb-821c-e29eadb1fe21.png)|![Screen Shot 2020-11-10 at 11 10 02 AM](https://user-images.githubusercontent.com/1202812/98699689-66d6c980-2345-11eb-826a-5e2f6abd3cc5.png)

**Tested with:**

[Seedlet Blocks](https://github.com/Automattic/themes/tree/trunk/seedlet-blocks), [Twenty Twenty-One Blocks](https://github.com/WordPress/theme-experiments/tree/master/twentytwentyone-blocks) 
Gutenberg 9.3.0
Safari 14.0

jasmussen:
Thanks for the report! Fix in #156.

johngodley:
The fix has been deployed

jasmussen:
:milkshake:

PR #156, opened 2020-11-16
Fix layout grid in site editor.

jasmussen:
This PR fixs #154 where the layout grid did not work in the site editor. The rules were simply not output, because https://github.com/Automattic/block-experiments/commit/4ac3fb6166024cc447725b8c5a5883b4cadd2c89 targets a rule that exists only in the post editor.

This PR hopes to keep the intent of that commit in place — scoping the rules only to the block editor — but target a class that exists in both the site and post editor.

Before:

Screenshot 2020-11-16 at 09 39 52

After:

Screenshot 2020-11-16 at 09 46 02

### DIFF
--- a/blocks/layout-grid/src/grid.scss
+++ b/blocks/layout-grid/src/grid.scss
@@ -49,7 +49,7 @@
 }
 
 // Editor grid styles
-.block-editor-page .wp-block-jetpack-layout-grid {
+.wp-block-jetpack-layout-grid {
 	// These three rules are only necessary for Safari.
 	// Safari interprets percentages in CSS grid differently.
 	// Flex causes children to "stretch", fixing that interpretation.
--- a/blocks/layout-grid/src/grid.scss
+++ b/blocks/layout-grid/src/grid.scss
@@ -49,7 +49,7 @@
 }
 
 // Editor grid styles
-.block-editor .wp-block-jetpack-layout-grid {
+.block-editor-page .wp-block-jetpack-layout-grid {
 	// These three rules are only necessary for Safari.
 	// Safari interprets percentages in CSS grid differently.
 	// Flex causes children to "stretch", fixing that interpretation.


johngodley:
Is `block-editor-page` always going to be on the page? The change in this PR looks fine, but it will break the block on P2. If `block-editor-page` is the thing to use then it can be added to P2.

jasmussen:
Good point.

But also per the conversation in https://github.com/WordPress/gutenberg/pull/26958, this issue may resolve itself.

johngodley:
> But also per the conversation in WordPress/gutenberg#26958, this issue may resolve itself.

The PR was merged (not sure which version) - is this PR still needed?

Addison-Stavlo:
> The PR was merged (not sure which version)

It doesn't look like the part that adds the `.block-editor` class to the Site Editor was ever merged however.

>  is this PR still needed?

Yes, I think this is still needed.  There is still reluctance to add that class to the site editor and until this PR (or that class) lands this will have a broken layout on dotcom FSE sites.

Addison-Stavlo (approved):
Tested and verified the class name change fixes the layout when used in the site editor.

johngodley:
> It doesn't look like the part that adds the .block-editor class to the Site Editor was ever merged however.

I think this needs some investigation. The problem with merging this PR is that it breaks it for P2, and we end up in a whack-a-mole situation.

Would it be a bad thing to apply it to `.block-editor` and `.block-editor-page` so it catches all setups?

Addison-Stavlo:
> Would it be a bad thing to apply it to .block-editor and .block-editor-page so it catches all setups?

What happens if we exclude both of them and have no global classname (https://github.com/WordPress/gutenberg/pull/27724#issuecomment-747351361)?   If it works without them that seems the way to go.

If not, I guess covering both setups this way makes sense.  It would be nice to have the `block-editor` class in the Site Editor as well (I had opened a tiny PR here to discuss - https://github.com/WordPress/gutenberg/pull/27724), but it seems some suggestions lean towards getting rid of that classname in all editors.  🤔

johngodley:
> What happens if we exclude both of them and have no global classname

A good point, and I'm not sure - it may be historical. @jasmussen can you remember?

jasmussen:
Slowly eeking my brain back to this PR after the new year break. Happy new year by the way!

> What happens if we exclude both of them and have no global classname

Gooooood question. Looking back at this, and seeing it in context, I am 99% sure it's to make it editor-only, as opposed to also outputting these rules on the frontend. The rules inside the wrapper include `> .wpcom-resize-grid >` which only appears in the editing interface. 

So it seems like another solution might be to move those rules to the editor.scss file? I'm just about to step away for the day, but folks should feel free to push to this branch, if such a change fixes it and works well. Thank you.

Addison-Stavlo:
> So it seems like another solution might be to move those rules to the editor.scss file? 

It looks like that is where that file is imported to. 👍  So theoretically should be safe to remove.  Will see if I can test it out in a bit.

Addison-Stavlo:
removed the extra class altogether - 
33fa871 - it seems to work well in the editor without changing the front-end, but could use a second check.

jasmussen:
Looks good to me. John what do you think?

david-szabo97 (approved):
Only the `front.scss` is loaded on the frontend. So this should be all right.

jasmussen:
Awesome, let's land it.

johngodley:
Yes, it doesn't have any negative effect on a standard theme or P2 that I can see.

I will release once another issue has been looked at.

Addison-Stavlo:
> I will release once another issue has been looked at.

Pardon the confusion, but has this happened yet?   I noticed the commit to bundle version 1.4 last week (https://github.com/Automattic/block-experiments/commit/2aa17beb1c6c0e8c482a67ad98ea1fb883020df2) - but looking at WordPress.org/plugins the Layout Grid Block still shows version 1.3.

johngodley:
@Addison-Stavlo it is still undergoing testing.

Addison-Stavlo:
Alright, thanks for confirming! 😁

PR #159, opened 2020-12-10
Add a hyphen, in those browsers that support it, to long words that are broken across lines

glendaviesnz:
Currently long words break between lines to avoid breaking layout. This PR adds a hyphen were possible to make it more obvious that this has happened.

#### Testing

- Check out this PR to local env
- Add a two column layout grid block with a long/large word and check that a hyphen is added if the word breaks across lines.

Before:
Screen Shot 2020-12-10 at 4 23 14 PM

After:
Screen Shot 2020-12-10 at 4 21 26 PM


jasmussen (approved):
Great improvement! Thank you!

PR #174, opened 2021-03-04
Try: Fix firefox boundary issue.

jasmussen:
Fixes https://github.com/Automattic/wp-calypso/issues/47247, hopefully with no side effects.

It effectively removes the fix we added in https://github.com/Automattic/block-experiments/pull/38, and it seems to work for me. Editor:

Screenshot 2021-03-04 at 10 22 47

Frontend:

Screenshot 2021-03-04 at 10 23 24

However it's important we test this in Firefox thoroughly so we don't regress the initial issue.

### DIFF
--- a/blocks/layout-grid/front.css
+++ b/blocks/layout-grid/front.css
@@ -1043,9 +1043,5 @@
   .wp-block-jetpack-layout-grid.wp-block-jetpack-layout-gutter__huge {
     grid-gap: 48px; }
 
-@-moz-document url-prefix() {
-  .wp-block-jetpack-layout-grid .wp-block-cover {
-    max-height: 0; } }
-
 .wp-block-jetpack-layout-grid-column {
   max-width: 100%; }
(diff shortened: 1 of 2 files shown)


jeyip:
Fixes https://github.com/Automattic/block-experiments/issues/176

johngodley:
This seems to work fine in Firefox.

Testing the original Codepen where I'd isolated the problem in Firefox 86.0 also seems to now work: 

https://codepen.io/johngodley/pen/vYOYxJe

The original problem page also looks fine (without this PR):

https://amwpapptesting.wordpress.com/tiled-gallery-and-other-testing/

It shouldn't affect other browsers, so it seems like we're good to go here.

Unfortunately, it now seems like Chrome (89.0.4389.72) has a similar problem (on the above URL):

image

Firefox and Safari both work fine. Chrome displays the codepen properly, so the cause is different, even if the end result is the same.

I don't know whether it's worth investigating the Chrome issue further. I wonder if this is something that could keep appearing and disappearing in browsers as the grid system matures. Also, the problem page also seems to contain unnecessarily complicated HTML. What do you think @jasmussen?

As the Chrome problem exists without this PR I'll merge this as it fixes #176.

jasmussen:
Hmm that's a doozy. And thank you for the fixes. I'll take a look.

PR #175, opened 2021-03-04
Remove hyphenation.

### LINKED ISSUE #155: Layout Grid Block: word wrap breaks words on smaller screens if the font-size is too big
#### Steps to reproduce the issue

1. Create a Layout Grid Block
2. Add an H1 heading, use the biggest word you can think off :) 
3. Save and check on very small resolution such of an iphone 5 or similar.

Related issue: https://github.com/Automattic/themes/issues/2666

#### What I expected

I expect to see the whole word on one line.

#### What happened instead

**Screenshots**

### LINKED ISSUE #173: Layout block: Hyphens Mobile Safari
### Steps to reproduce the behavior

1. Add a Layout Block to a page. 
2. Inside of this block, add a heading block with a long text.  Here is an example - "My name is James and I am a Happiness Engineer at Automattic."
3. Publish the page a view it in Safari on mobile. 

#### What I expected to happen
On other mobile browsers, words are breaking with no hyphens.

#### What actually happened
In Safari, the CSS is currently set to `-webkit-hyphens: auto;` when I believe it should be set to `-webkit-hyphens: none;`.  

### Context
3702064-zen

#### Browser / OS version

Model: iPhone (seems to affect any model)
Browser:Safari
IOS: 13.1.3 (this is the current version that is available in Safari's **Responsive Design Mode**

#### Is this specific to the applied theme? Which one?
I can reproduce this on any theme I have tested so far.  The CSS is coming from the block itself.  

#### Does this happen on simple or atomic sites or both?
Both

#### Is there any console output or error text?
None

#### Level of impact (Does it block purchases? Does it affect more than just one site?)
This affects all users in Safari that are currently using this block.  We can provide the following CSS to temporarily resolve this for the user.

```
.site-content .wp-block-jetpack-layout-grid .wp-block-jetpack-layout-grid-column *{
    -webkit-hyphens: none;
}
```

#### Screenshot / Video: If applicable, add screenshots to help explain your problem.
Markup on 2021-02-06 at 14:36:36

jasmussen:
Fixes #173, but undoes the fix we used to close #155. 

Text wraps so as to not break the layout, see https://github.com/Automattic/block-experiments/issues/173#issuecomment-790529713.

However hyphenation has not been popular. This PR removes it again.

### DIFF
--- a/blocks/layout-grid/front.css
+++ b/blocks/layout-grid/front.css
@@ -1030,10 +1030,7 @@
             grid-row-start: 4; } }
   .wp-block-jetpack-layout-grid .wp-block-jetpack-layout-grid-column * {
     word-break: break-word;
-    word-wrap: break-word;
-    -webkit-hyphens: auto;
-        -ms-hyphens: auto;
-            hyphens: auto; }
+    word-wrap: break-word; }
   .wp-block-jetpack-layout-grid.wp-block-jetpack-layout-gutter__none {
     grid-gap: 0px; }
   .wp-block-jetpack-layout-grid.wp-block-jetpack-layout-gutter__small {
--- a/blocks/layout-grid/front.scss
+++ b/blocks/layout-grid/front.scss
@@ -106,7 +106,6 @@
 	.wp-block-jetpack-layout-grid-column * {
 		word-break: break-word;
 		word-wrap: break-word;
-		hyphens: auto;
 	}
 
 	&.wp-block-jetpack-layout-gutter__none {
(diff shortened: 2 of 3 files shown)


mikeicode:
@jasmussen This appears to remove the hyphens, and it looks like this change was committed. 

The hyphens still exist however, from a Simple Site:

https://d.pr/i/xIIFsn

Should this be live on all sites now? We are still receiving support requests to remove them which is why I'm asking and am confused by the status of this change. Thanks!

johngodley:
It hasn't been deployed yet. It will be once it has been reviewed for WordPress.com

PR #177, opened 2021-03-09
Remove front.css

### LINKED ISSUE #170: Layout Grid: Doesn't work with block based themes
Layout Grid doesn't work with block-based themes. `front.css` isn't loaded on the front at all.

Looks like we are registering the stylesheet on `wp_enqueue_scripts` action:

https://github.com/Automattic/block-experiments/blob/7ca7c4d20c45699069a1fe3b3d1fd3e2e2423710/blocks/layout-grid/index.php#L29-L34

We should consider moving this into `init` right before `register_block_type( 'jetpack/layout-grid', [`.

johngodley:
This moves all the CSS from `front.css` into `style.css`. The block then operates as a normal block without any additional loading of CSS files.

The reason a `front.css` was created is to better seperate the grid styles in the editor from the ones on the front. These are necessarily different (the editor has a very different DOM to the rendered block), but were causing clashes.

To work around this I've tied all the editor grid CSS to `wp-block-jetpack-layout-grid-editor` instead of `wp-block-jetpack-layout-grid`.

This should mean it will now work with block-based themes.

Fixes #170


johngodley:
I think it looks worse than it is. If problems are going to surface it will be in the editor, not the front end, so it should be fairly safe in terms of not breaking a site.

jasmussen:
Let's do it.

mkaz:
Confirmed this fixed the issue with a block-based theme and layout grid not working on front-end. :+1:

PR #183, opened 2021-03-30
Layout Grid: Use desktop layout when rendering within a pattern preview (e.g. within a Disabled context)

andrewserong:
Related to: 234-gh-Automattic/view-design

This change adds a HOC to load in the Disabled context for the Layout Grid to determine whether or not we are rendering within a block pattern preview. If we are, then let's render using the desktop layout, to prevent very long vertical pattern previews (see: 234-gh-Automattic/view-design)

After this change, the block pattern preview will render using the Desktop layout when you're using a mobile viewport, however once you insert the pattern, you will be _editing_ using the mobile viewport, so that 🤞 we get the best of both worlds.

| Mobile | Desktop |
| --- | --- |
| image | image |

How to test:

Testing this with the wpcom patterns is a little tricky, but I managed to get it working using a8c-wp-env (loading the ETK plugin and Block Experiments, and then going to the Plugins to install Gutenberg), and then adding the following line to `block-experiments.php` so that we switch on loading the wpcom patterns for testing:

```
add_filter( 'a8c_enable_block_patterns_api', '__return_true' );
```

I also needed to run `npm build` in the block experiments directory to get this running.

Once your local development environment is running this change correctly, test the following:

1. When in a desktop viewport, open up the pattern inserter and the first pattern under Featured should be rendered using the Desktop layout within the pattern preview. When you insert the pattern, it should still be in the Desktop layout.
2. Using a mobile viewport (e.g. narrow your browser window), when you open up the pattern inserter, the first pattern under Featured should still be rendered using the Desktop layout within the pattern preview. However, when you insert it, it should now use the mobile layout while you're editing your post or page.

### DIFF
--- a/blocks/layout-grid/src/grid/edit.js
+++ b/blocks/layout-grid/src/grid/edit.js
@@ -211,6 +211,11 @@ class Edit extends Component {
 	}
 
 	getPreviewMode() {
+		// If we're rendering within a pattern preview, use the desktop layout for the preview.
+		if ( this.props.isBlockOrPatternPreview ) {
+			return 'Desktop';
+		}
+
 		// If we're on desktop, or the preview is set to mobile, then return the preview mode
 		if (
 			this.state.viewPort === 'Desktop' ||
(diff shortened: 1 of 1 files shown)


johngodley:
Works for me, but looping @jasmussen in for any additional thoughts.

jasmussen:
Thanks for the PR. Let me be sure I understand correctly, that this change affects _only_ the preview of the pattern in the inserter?

If yes, then that seems okay, especially with John's blessing. 

Though do note that at some point in the future, responsiveness (whether classic or [container queries](https://twitter.com/una/status/1375419967718449155)) will land in the core project, and we will likely want something similar to exist there. But I guess that's a bridge we can cross at that point in time.

johngodley:
I will include this in the 1.6 release

andrewserong:
Thanks, folks! Yes, this is just for the preview in the inserter. Much appreciated.

Also, thanks for the link about container queries, I didn’t realise they were available to play with in Chrome, sounds super exciting 😀

jasmussen:
If we can get Safari and Firefox to adopt them, container queries would be an ideal solution for providing responsive features to blocks, and perhaps especially to patterns.

PR #185, opened 2021-04-12
Try: Padding fix.

jasmussen:
https://github.com/Automattic/block-experiments/pull/177 retired the front only stylesheet.

It also changed `.wp-block-jetpack-layout-grid` to `.wp-block-jetpack-layout-grid-editor`, in the editor only. Which appears to have caused some issues with custom column paddings, and grid end gutters as well.

This PR does fix that:

patch

However I'm not sure this is the way we should fix it. In part also because I'm not entirely sure _why_ it fixes it. A better fix might be to not rename that class in the first place.

### DIFF
--- a/blocks/layout-grid/style.scss
+++ b/blocks/layout-grid/style.scss
@@ -10,7 +10,7 @@
  */
 
 /* autoprefixer grid: no-autoplace */
-
+.wp-block-jetpack-layout-grid-editor,
 .wp-block-jetpack-layout-grid {
 	// This padding adds end gutters.
 	padding-left: 24px;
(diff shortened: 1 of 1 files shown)


johngodley:
This fixes the end gutters when I try in 2021. I wasn't able to see a problem with the gutter size setting, but maybe it's theme specific.

> A better fix might be to not rename that class in the first place.

The problem we have is that the editor DOM and front-end DOM are different, and so different styles are needed to apply the grid. However, because we use the same class names then the front-end grid is partially affecting the editor grid in unpredictable ways.

The simplest solution seemed to be found in renaming the editor class so it's not affected by the front end.

I agree that it's not ideal. I would hope that a lot of the need for this will disappear with #105, but that's a much bigger job that requires more work. I was hoping the Global Grid would appear in the meantime and negate the need for it!

I'll look this into the wp.com diff and try it out on a few more themes, but otherwise it's looking good.

johngodley:
Confirmed that this also fixes the individual column padding, which I missed. Let's merge and take this over to wp.com

PR #193, opened 2021-05-13
Promote duotone to core

ajlende:
Duotone was merged into WordPress/gutenberg! 🎉 

The new source lives in Gutenberg, so the code here can be removed and replaced with a link to its new home.


jasmussen:
Nice! Ship it.

PR #196, opened 2021-05-25
Remove duotone.

jasmussen:
It looks like not all duotone had been removed in #193. This removes the rest.

PR #201, opened 2021-06-02
Remove duotone from experiments

johngodley:
It migrated to core!


jasmussen:
This looks remarkably similar to https://github.com/Automattic/block-experiments/pull/196, how is that possible? 😅 🙈

johngodley:
Git is still a ⬛ !

PR #208, opened 2021-06-15
Initial commit of experimental Draft block

glendaviesnz:
### Description
One of the criticisms of the Block editor, particularly from those that are used to the Classic editor, is that it has a distracting interface for long form writing. While it is possible to reduce the noise by pinning the toolbar to the top, there are still issues with editing compared to a standard text editor, particularly if trying to copy segments of text that span paragraphs, ie. copying the last sentence of one paragraph and the first of another:

While this may not be common when editing basic web pages, I can see how this would be a very common annoyance for long form writing.

The intention of this new 'Draft' block is to allow for easy writing/editing of long form articles with only very basic text formatting options, and conversion of the content to paragraph blocks once editing is finished to allow for more complex layout and formatting at that point:

The advantage of this approach is that basic writing and text editing is much more fluid and similar the the usual text editor experience.

Some more background discussion at pcjTuq-nG-p2

### DIFF
--- a/blocks/draft/editor.scss
+++ b/blocks/draft/editor.scss
@@ -1,0 +1,10 @@
+.wp-block-jetpack-draft {
+	.draft-label {
+		text-align: right;
+		font-size: 0.8rem;
+	}
+	.draft-content {
+		border: 1px dashed #cccccc;
+		padding: 8px;
+	}
+}
--- a/blocks/draft/index.php
+++ b/blocks/draft/index.php
@@ -1,0 +1,18 @@
+<?php
+function draft_block_init() {
+	register_block_type( 'jetpack/draft', [
+		'editor_script' => 'block-experiments',
+		'style' => 'block-experiments',
+		'editor_style' => 'block-experiments-editor',
+		'render_callback' => 'render_block_draft',
+	]);
+}
+add_action( 'init', 'draft_block_init' );
+
+function render_block_draft( $attributes, $content ) {
+	if ( ! current_user_can( 'editor' ) && ! current_user_can( 'administrator' ) ) {
+		return;
+	}
+	$draft_label = '<div class="wp-block-jetpack-draft__label">' . __( 'Draft content: only visible to editors' ) . '</div>';
+	return $draft_label . $content;
+}
--- a/blocks/draft/readme.md
+++ b/blocks/draft/readme.md
@@ -1,0 +1,19 @@
+## Draft content editing block
+
+This is currently just an experimental block that allows easier drafting of post text content, which can then be converted to blocks for more complex formatting once the text is finalised.
+
+### Why?
+
+One criticism of the block editor over the old classic editor is that it can be a distracting enviroment for those that are drafting long form posts as apposed to contructing website pages. Although the level of distraction can be reduced by pinning the block toolbar to the top, copying and pasting, particularly getting partial text between two paragraph blocks, is not as fluid and easy as in a standard text editor.
+
+This block allows very quick and easy text editing, with basic text formatting, for initial post writing, which can then be converted to standard paragraph blocks to allow all the flexibility of the block editor for final formatting.
+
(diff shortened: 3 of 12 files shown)


jasmussen:
Nice work. This is what I see:

drafts 2

The primary win is that the partial text-selection affordances you see in text editors is available here. I miss contacting two paragraphs into one too at times, and when I do, I jump into my separate Markdown editor, knowing that copy/paste just shines in the block editor. In fact, I'd love to see partial selections come to the vanilla paragraph blocks directly, a feat that seems within reach if given the effort.

That's also where I think this block might have a challenge to. Partial selections are nice, but on their own I'm not sure they are worth the overhead of an entirely separate block having to be inserted first. I'm also not sure I agree this block makes for a less distracting writing flow than vanilla paragraphs: the "Finalize content" button has a fair bit of prominence, there's a permanent notice for editors, and the dashed line around the block all serve to add visual noise and depart from WYSIWYG where paragraph blocks do not. 

"Draft" is a great name, though. Are there other features it could offer in addition to partial selections that might lean into that concept?

glendaviesnz:
Thanks for the feedback @jasmussen, all good valid questions - got tied up today with other things, but will try and respond tomorrow.

ramonjd:
Just took this for a spin. It's working very well!

A few ideas occurred to me:

Would it be useful to detect list bullet points and transform lists when finalizing? Or maybe even auto-adding the bullet points, similar to how Github textinputs do?

Screen Shot 2021-06-17 at 10 14 31 am

This is just a personal observation as I do make a lot of lists when drafting notes :D

My favourite functionality is that I can add content to a post, publish but still have that content hidden on the frontend. I'm thinking that it would be interesting to extend the Draft block to be able to house any block/pattern - sort of like a Group block but for drafts 😄 

On the frontend, perhaps it would be handy to indicate the areas which are drafts and not published to a logged-in editor:

Screen Shot 2021-06-17 at 10 21 50 am

But then I realized that it would be equally handy to see how the content would look without such indicators. 

Not sure, just a brain dump here. Thanks!

jasmussen:
> My favourite functionality is that I can add content to a post, publish but still have that content hidden on the frontend. I'm thinking that it would be interesting to extend the Draft block to be able to house any block/pattern - sort of like a Group block but for drafts 😄

It needs to be clear why it's not on the frontend, and how you make it shown. But I think you're on to something here, it's leans into the "Draft" idea and provides a niche. What other tools does it enable? While I'm still overall skeptic of broad adoption, embracing this flow could make it just the block for a particular group of people.

glendaviesnz:
Thanks for the feedback @jasmussen and @ramonjd, and sorry about the delay in responding.

>the "Finalize content" button has a fair bit of prominence, there's a permanent notice for editors, and the dashed line around the block all serve to add visual noise and depart from WYSIWYG where paragraph blocks do not.

The reasoning behind these bits of additional UI was an initial thought that we may not want to see the block used as the final block on the page - the risk being that people would then start wondering why they can't add images, or more complex layout once they have finished writing the text. Maybe there would be better UX to indicate this. 

Or as you also note, maybe the the copying and pasting of text between paragraphs can be handled better within the editor, in which case, maybe the focus should be on the other thing this block brought to light which is the ability to add draft content anywhere within a new or existing page.

>My favourite functionality is that I can add content to a post, publish but still have that content hidden on the frontend. I'm thinking that it would be interesting to extend the Draft block to be able to house any block/pattern - sort of like a Group block but for drafts

Yeh I think this could be a really useful block, but I think it is something very different to what this initial draft block was intended, but could be created as a completely separate block as you not, or as some extensions to the Group block.

Will give all this some further thought before going any further with this.

glendaviesnz:
Given that this approach goes against some of the initial deliberate design decisions in the editor (pcjTuq-nG-p2) I am just going to close this for now. It has been useful for exposing some ideas about a draft content option if nothing else.

PR #226, opened 2021-07-07
Layout Grid: Correctly align image block with right alignment

mikachan:
Currently, when adding an image in a Layout Grid block and then setting the image alignment to the right, the change is not reflected in the editor. It works fine on the front end. This PR fixes this by applying `float: right` to the image container (similar to the styles for alignment on the front end.)

image

#### Testing steps:
- Add a layout grid block
- Add an image to the block
- Align the image to the right
- The image should align to the right of its column in the editor

This also addresses a [related issue in themes](https://github.com/Automattic/themes/issues/4158).

### DIFF
--- a/blocks/layout-grid/editor.scss
+++ b/blocks/layout-grid/editor.scss
@@ -246,3 +246,13 @@
 		}
 	}
 }
+
+/**
+ * Image alignment for the editor
+ */
+
+.wp-block[data-align=right] {
+    > .wp-block-image {
+        float: right;
+    }
+}


johngodley:
Looping in @jasmussen as I have no real opinion about this.

jasmussen:
Thanks for the ping and thanks for the PR. I'll definitely take a look tomorrow morning my time, and see if we can merge this one!

jasmussen:
Thank you for the PR, I took it for a spin, first confirming that yes, if you right align an image block inside a layout grid column it's right aligned on the frontend but not inside the column. Also, that this PR fixes that:

Screenshot 2021-07-08 at 10 37 57

Screenshot 2021-07-08 at 10 38 02

I thought this quite curiuos; surely a right aligned image should just be right aligned regardless of context, right? Well as it turns out, this set of rules defines right alignments:

Screenshot 2021-07-08 at 10 39 59

That is, there are separate rules for top level blocks and nested blocks. For example, rules for floats inside groups are unique:

Screenshot 2021-07-08 at 10 43 06

Screenshot 2021-07-08 at 10 41 55

All of this appears to be specific to block themes that declare `layout` in theme.json, so I suspect it'd work fine in classic themes. But I also feel like this is a problem that should be fixed directly in the block editor. I.e. instead of editor rules that look like this:

```
.editor-styles-wrapper .wp-container-17 > [data-align="right"] {
    float: right;
    margin-left: 2em;
}
```

could we simply have this?

```
.editor-styles-wrapper .wp-block > [data-align="right"] {
    float: right;
    margin-left: 2em;
}
```

?

That would theoretically scale to more plugin blocks. 

Technically I'd be happy to land this PR as is, and as a bandaid. But it might be worth it to still look at an upstream PR to explore fixing it at the source, or at least figure out why nested block floats have such unique generated slugs. I think I recall Riad working on `layout` specific changes, he might be good to ping on a PR.

What do you think, @mikachan? And thanks again for the PR!

mikachan:
Thanks so much for taking a look @jasmussen!

This is a great insight into how these styles are being applied. I didn't realise they were so specific to block themes that declare `layout` in theme.json. Perhaps this would be more suited to a fix in the block editor. I'll take a look today and report back here. Thanks so much for your input and suggestions, it's really helped with my understanding here.

jasmussen:
Always happy to help 👌

mikachan:
After taking a look in the block editor and trying a few different solutions out, I'm not sure which one is best so I've come back for some advice, if that's OK!

The CSS you mentioned seems like the ideal place for this fix:

```
.editor-styles-wrapper .wp-block > [data-align="right"] {
    float: right;
    margin-left: 2em;
}
```

However, this style is currently applied via a component called [LayoutStyle](https://github.com/WordPress/gutenberg/blob/7043eaa2ba45a9d418f5c33a10d3e119fffb94bc/packages/block-editor/src/hooks/layout.js#L216), which applies these properties to selectors that are passed in through to the component. It looks like the Layout Grid block isn't using this component, as the inline styles that this component adds aren't applied when changing the alignment, as they are when you change alignment in a group block. To confirm, the `style` tag isn't even added to the DOM for these blocks as it is with the group block.

I don't have much experience with Gutenberg so I'm not sure how we'd apply these styles using this component, at a guess I'd say we possibly need to wrap the Layout Grid in this component? I'm happy to investigate myself if you think this is the right solution.

Another option is to add the above style block to the [editor](https://github.com/WordPress/gutenberg/blob/trunk/packages/block-library/src/editor.scss) CSS in Gutenberg (or another CSS file). This works, but it doesn't feel right as we're then repeating these styles, although they are already repeated a couple of times. I understand this is probably more of a discussion to have in the Gutenberg repo.

And then another option is to apply the same CSS in this repo, and perhaps make it more specific with a Layout Grid class so it only affects this block.

I'm happy to ask some/all these questions over in Gutenberg. I just wanted to report back what I'd found, as it doesn't seem as straightforward as adding these 2 lines of CSS - although I think that's an option, I'm just not sure where!

scruffian:
I agree, let's open an issue Gutenberg, but also merge this as a bandaid. Thanks @jasmussen

jasmussen:
👍 👍 go for it!

johngodley:
I'll include this in the next release.

scruffian:
Thanks @johngodley

mikachan:
Thanks so much all for your help, and for including this quick fix in the next release! I've opened an issue in Gutenberg: https://github.com/WordPress/gutenberg/issues/33332

PR #233, opened 2021-07-20
Fix styling for center-aligned events

ajlende:
Fixes an issue with center-aligned event blocks, and fixes block directory details.

Before:

Screen Shot 2021-07-20 at 14 12 50

After:

Screen Shot 2021-07-20 at 14 13 49

### DIFF
--- a/blocks/event/editor.scss
+++ b/blocks/event/editor.scss
@@ -10,7 +10,9 @@
 	}
 
 	// Fixes for overriding editor styles
-	.editor-styles-wrapper & {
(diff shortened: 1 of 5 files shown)


Aurorum:
@ajlende I received a bug report about the Events Block in 7123218-zen. You appear to have been the last person to have contributed to it, so would you have any idea if the bug is worth formally reporting or is this block no longer maintained? Can't find any reference to it anywhere, but it's still on the Plugin Directory - to be honest, I didn't know Automattic did this block!

Thanks! :)

PR #234, opened 2021-07-20
Use new frontend scripts method for waves

ajlende:
Uses [`should_load_separate_core_block_assets`](https://make.wordpress.org/core/2021/07/01/block-styles-loading-enhancements-in-wordpress-5-8/) instead of `render_callback` for loading the frontend script only when the block is loaded.

Also loads script translations with `wp_set_script_translations` since that seems to have been missing when they were added to the other blocks.

Tested with 5.8.

### DIFF
--- a/blocks/waves/index.php
+++ b/blocks/waves/index.php
@@ -2,13 +2,10 @@
 
 add_action( 'init', function() {
 	register_block_type( 'a8c/waves', [
+		'script' => 'a8c-waves-js',
 		'editor_script' => 'block-experiments',
 		'style' => 'block-experiments',
 		'editor_style' => 'block-experiments-editor',
-		'render_callback' => function( $attribs, $content ) {
-			wp_enqueue_script( 'a8c-waves-js' );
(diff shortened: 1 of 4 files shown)


blocks/waves/index.php
```diff
@@ -24,8 +21,7 @@
 		filemtime( plugin_dir_path( __FILE__ ) . 'waves.js' ),
 		true // in footer
 	);
+	wp_set_script_translations( 'block-experiments', 'waves' );
 } );
 
-add_action( 'enqueue_block_editor_assets', function() {
-	wp_enqueue_script( 'a8c-waves-js' );
-} );
+add_filter( 'should_load_separate_core_block_assets', '__return_true' );
```
gziolo, line 27: I don't think that an individual block should enable this flag because it's something that should be reserved mostly for themes or site owners. Se the full dev note for more context: https://make.wordpress.org/core/2021/07/01/block-styles-loading-enhancements-in-wordpress-5-8/.

PR #246, opened 2021-11-09
Layout Grid: Add support for the Layout Grid to the excerpt

### LINKED ISSUE #245: Layout Grid: Contents in the Layout Grid block are being excluded from the excerpt
This has been originally reported here: https://github.com/Automattic/wp-calypso/issues/54046

I was able to track the issue down to the `excerpt_allowed_wrapper_blocks` array.

I don't know if the Layout Grid contents are being intentionally excluded from the excerpt generation - if that's not the case, we should probably allow it, as it is confusing users.

`'jetpack/layout-grid', 'jetpack/layout-grid-column',` should be added to the list via `excerpt_allowed_wrapper_blocks` filter.

danieldudzic:
Closes #245

### DIFF
--- a/blocks/layout-grid/index.php
+++ b/blocks/layout-grid/index.php
@@ -16,3 +16,7 @@ add_action( 'init', function() {
 
 	wp_set_script_translations( 'block-experiments', 'layout-grid' );
 } );
+
+add_filter( 'excerpt_allowed_wrapper_blocks', function( $allowed_wrapper_blocks ) {
+	return array_merge( $allowed_wrapper_blocks, array( 'jetpack/layout-grid', 'jetpack/layout-grid-column' ) );
+} );


jasmussen:
Thanks for the PR. As a quick and very basic test, this works as intended:
Screenshot 2021-11-10 at 08 55 09
Screenshot 2021-11-10 at 08 54 49

I assume this addresses [Velda's concern about RSS feeds breaking](https://github.com/Automattic/block-experiments/issues/245#issuecomment-964367687)? 

Code looks good, if you have good confidence in this one, ship it.

danieldudzic:
@jasmussen Thank you for the review. Could you please provide guidance on how to best handle shipping this?
Not sure who is responsible for the block experiments repo and also what the process of pushing updates is. Thank you 🙏

johngodley:
I can release this.

PR #247, opened 2021-11-17
Full Site Editing: Prefer site editor viewport settings when changing the preview device type

zaguiini:
Fixes https://github.com/Automattic/wp-calypso/issues/57158.

When we are trying to set a different preview device inside FSE, the "Layout Grid" block ignores the FSE device type and uses the internal (`core/edit-post`) responsive breakpoints setting instead of `core/edit-site`, which should take precedence inside FSE.

We must take the `core/edit-site` registry into consideration, which is only registered if we are inside FSE, to properly layout the grid when changing the device type.

Ideally, we would have a centralized, agnostic way of referencing/setting the preview device type, instead of needing to reference two different registries. I'm happy to create an issue as a follow-up, as it possibly demands extensive refactor.

## Behavior on `master`

https://user-images.githubusercontent.com/10015451/137959268-8e3b4189-3278-427a-9441-0cb265cc26b3.mp4

## Expected/this branch's behavior

https://user-images.githubusercontent.com/10015451/137959354-99eb95b3-8864-4538-a326-04549fd50314.mp4

Check that it works both on post and site editor.

### DIFF
--- a/blocks/layout-grid/src/grid/higher-order.js
+++ b/blocks/layout-grid/src/grid/higher-order.js
@@ -122,9 +122,9 @@ export function withSetPreviewDeviceType() {
 	return withDispatch( ( dispatch ) => {
 		return {
 			setPreviewDeviceType( type ) {
-				const { __experimentalSetPreviewDeviceType } = dispatch(
-					'core/edit-post'
-				);
+				const { __experimentalSetPreviewDeviceType } =
+					dispatch( 'core/edit-site' ) ||
+					dispatch( 'core/edit-post' );
 				__experimentalSetPreviewDeviceType( type );
 			},
 		};
@@ -158,9 +158,8 @@ export function withColumnAttributes() {
 
 export function withPreviewDeviceType() {
 	return withSelect( ( select ) => {
-		const { __experimentalGetPreviewDeviceType = null } = select(
-			'core/edit-post'
-		);
+		const { __experimentalGetPreviewDeviceType = null } =
+			select( 'core/edit-site' ) || select( 'core/edit-post' );
 
 		return {
 			previewDeviceType: __experimentalGetPreviewDeviceType(),


blocks/layout-grid/src/grid/higher-order.js
```diff
@@ -122,9 +122,9 @@ export function withSetPreviewDeviceType() {
 	return withDispatch( ( dispatch ) => {
 		return {
 			setPreviewDeviceType( type ) {
-				const { __experimentalSetPreviewDeviceType } = dispatch(
-					'core/edit-post'
-				);
+				const { __experimentalSetPreviewDeviceType } =
+					dispatch( 'core/edit-site' ) ||
+					dispatch( 'core/edit-post' );
```
Addison-Stavlo, line 127: Just a quick note that this should work for us given the order (both here and in the selects further down in the file).  `core/edit-post` is available in the Site Editor, but `core/edit-site` is not in the post editor.  So as long as that functionality stay the same, this should work as expected.

It's been like this for a while and I don't expect it to change soon, but I wonder if there is a more robust way to determine the editor in use in case it did. 🤔

zaguiini: > I wonder if there is a more robust way to determine the editor in use in case it did

That's what me and @jeyip thought, too. We couldn't find anything like that, but maybe there's more that we don't know. Leaving it to WP masters!

jasmussen:
Thanks for the PR. Just as a quick glance, the code looks good to me. But specifically I'd defer to the both of you to sanity check the inner functionality, and you should feel free to land this one when you have good confidence.

We'll need to make a release when it's all ready.

jeyip:
## Testing
### Requirements
- [x] Layout grid block behaves responsively for "tablet" preview device mode
- [x] Layout grid block behaves responsively for "mobile" preview device mode

### Browsers
- [x] Edge
- [x] Firefox
- [x] Safari
- [x] Chrome

jeyip:
> I wonder if there is a more robust way to determine the editor in use in case it did.

I 100% agree with this sentiment. We're definitely open to suggestions. Luis and I couldn't think of anything else after giving it more thought.

As far as the changes in this PR go, they look reasonable and they test well. Luis created [a P2](p58i-buY-p2) for thoughts about an alternative approach, and we'll also probably wanna wait a little longer for feedback about how to determine if the site editor is in use. Everything else lgtm!

jeyip (approved):

PR #252, opened 2021-12-15
Layout Grid: Fix vertical alignment and background colors.

### LINKED ISSUE #250: Layout Grid: column vertical alignment doesn't work in preview
It works fine on the live site. However, preview always displays the column's contents at the top, vertically speaking.

| Preview | Live site |
| -------- | -------- |
| image | image |

Notice how the center alignment option is selected in the Column block.

jasmussen:
Fixes #250.

The primary issue this solves is to ensure that the vertical alignment of content inside a column works in the _site editor_, where it's currently broken:

site editor is not

 In the post editor, it's fine:
post editor is fine

In exploring the cause and the fix, it appears that both markup and structure has changed and simplified a bit in the core offering, meaning some rules targetting specific markup stopped working. I also recall some upstream vertical centering rules changing — those rules may or may not affect this plugin. @andrewserong if you have time to look, I'd appreciate it. 

I also realized based on the commentary, that another feature had likely stopped working:

> // Make sure each column is full height in the editor, as it is on the frontend.

Specifically, I recall a use case being to apply background colors on a per-column basis. That appears to have broken, both in the post and site editors:

Screenshot 2021-12-15 at 11 08 00

This PR fixes it for both editors:
Screenshot 2021-12-15 at 10 59 46

Screenshot 2021-12-15 at 11 00 09

### DIFF
--- a/blocks/layout-grid/editor.scss
+++ b/blocks/layout-grid/editor.scss
@@ -35,6 +35,7 @@
  */
 
 // Make sure each column is full height in the editor, as it is on the frontend.
+[data-type="jetpack/layout-grid-column"],
 [data-type="jetpack/layout-grid-column"] > .block-editor-block-list__block-edit,
 [data-type="jetpack/layout-grid-column"] > .block-editor-block-list__block-edit > [data-block],
 .wp-block-jetpack-layout-grid-column {
--- a/blocks/layout-grid/style.scss
+++ b/blocks/layout-grid/style.scss
@@ -186,16 +186,22 @@
  * Individual column alignment
  */
 .wp-block-jetpack-layout-grid-column {
+	display: flex;
+	height: 100%;
+
 	&.is-vertically-aligned-top {
 		align-self: flex-start;
+		align-items: flex-start;
 	}
 
 	&.is-vertically-aligned-center {
 		align-self: center;
+		align-items: center;
 	}
 
 	&.is-vertically-aligned-bottom {
 		align-self: flex-end;
+		align-items: flex-end;
 	}
 }
 


blocks/layout-grid/editor.scss
```diff
@@ -35,6 +35,7 @@
  */
 
 // Make sure each column is full height in the editor, as it is on the frontend.
+[data-type="jetpack/layout-grid-column"],
 [data-type="jetpack/layout-grid-column"] > .block-editor-block-list__block-edit,
 [data-type="jetpack/layout-grid-column"] > .block-editor-block-list__block-edit > [data-block],
```
jasmussen, line 40: The `block-editor-block-list__block-edit` class appears to be gone from the editor. I've kept the two rules for back compat, and because they are harmless.

blocks/layout-grid/style.scss
```diff
@@ -186,16 +186,22 @@
  * Individual column alignment
  */
 .wp-block-jetpack-layout-grid-column {
+	display: flex;
+	height: 100%;
```
jasmussen, line 190: This height rule likely makes the editor-only height rule redundant, [this one](https://github.com/Automattic/block-experiments/pull/252/files#diff-479fde507612e22d5a3adf773cece3ab696340ff1f19b7ab771056be352b5de3R42). I still kept it for the time being.

blocks/layout-grid/style.scss
```diff
@@ -186,16 +186,22 @@
...
+
 	&.is-vertically-aligned-top {
 		align-self: flex-start;
+		align-items: flex-start;
```
jasmussen, line 194: It seems this is the key change that allows both a full height column, so the background color can stretch to fill the vertical space, and apply the alignment inside. Probably the reduction in containers made align-self need replacing by an align-items instead.

andrewserong (commented):
Thanks for the ping @jasmussen!

For my testing, prior to applying this PR, this is how a three column Layout Grid block looks for me on the front end of my test site:

image

With it applied, with the addition of `display: flex` the blocks within each column default to being laid out horizontally:

image

I had a go at adding `flex-direction: column`, which gets us closer to the correct layout. However, here I notice that switching the column to `display: flex` introduces the same issue as in https://github.com/WordPress/gutenberg/pull/32909 where due to the children of the column now being flex items, the margin collapse no longer applies, so the spacing between individual blocks is now larger than prior to this PR:

| Before | After |
| --- | --- |
| image | image |

I think that's the main reason why that PR never wound up landing if I remember correctly.

---

Getting back to the issue of vertical alignment. It looks like prior to this PR the vertical alignment is working correctly on the front end of the site, but not within the site editor? In the site editor, the column is stretched to be full height instead of the height of the content, so the vertical alignment doesn't appear to do anything within the editor. The furthest I could get in debugging this is that the `height: 100%` [rule in this block's editor styles](https://github.com/Automattic/block-experiments/blob/2fb575136e9b5f324b2d377764b3fbb1db1c6b37/blocks/layout-grid/editor.scss#L41) appears to behave differently in the post editor vs the site editor, but I couldn't quite work out why, and whether it's the presence or absence of styles in the site editor that's causing the difference 🤔

I'm not sure if any of that's helpful, sorry! But I think due to the margin collapse (or lack thereof) issue, I would probably lean toward looking into why `height: 100%` behaves differently between the two editing environments.

jasmussen:
This is incredibly helpful, thank you, I'm so happy I managed to ping the right person on this. That was indeed the conversation I vaguely recalled, and it seems clear from your feedback that I need to have another go at this one. I'll try and see if I can push a solution that introduces less big changes 👌 

I hope you won't mind if I ping you on the new PR also, when it's ready?

jasmussen:
I created #254 as an alternative to this one, it should be a bit smaller. What do you think?

andrewserong:
Thanks @jasmussen, I've just given #254 the ✅ 👍 

> I hope you won't mind if I ping you on the new PR also, when it's ready?

Always happy for you to ping me on things, and if it winds up being outside of my area I'll either read up to get the context or let you know 😁. I'll be AFK for a couple of weeks from the end of next week, but other than that, always happy to lend a hand!

jasmussen:
I appreciate that, thank you!

jasmussen:
Closing in favor of #254.

Issue #249, opened 2021-11-23
Layout Grid does not respect Group constraints


zaguiini:
## Description

When wrapping the Layout Grid with a Group so you can set layout constraints, the changes are not applied.

## Steps to reproduce

1. Insert Layout Grid on the page;
2. Wrap it within a Group;
3. Add Group layout constraints (e.g. column 500px wide 500px);
4. Save.

Check that the layout grid is still occupying the full width even though it was explicitly wrapped in a Group.

## Screenshots

![image](https://user-images.githubusercontent.com/26530524/143097794-4ee7bf0d-f7e3-4c91-bad0-d445888a44e0.png)

![image](https://user-images.githubusercontent.com/26530524/143097824-662a38b7-3587-4873-87cb-c50f33593543.png)

![image](https://user-images.githubusercontent.com/26530524/143097852-7c18cc0a-2f48-4b42-9ac3-afea07f7fa82.png)

jasmussen:
Thanks for the ticket. I can't reproduce this, the layout grid does inherit the constraints set on the parent group:
<img width="1061" alt="Screenshot 2021-12-17 at 09 39 29" src="https://user-images.githubusercontent.com/1204802/146515126-86fabf28-ad73-4440-bf1c-d726a3ec9320.png">

A small preface:

- When the group sets content and wide constraints, those constraints apply to the content inside (no position set) and the wide position.
- By default, the layout grid when inserted is set to full-wide. Full-wide items inside a group _should_ be the full width of the group, but no bigger.

Is it possible you tested with a full-wide layout grid inside a full-wide group?

zaguiini:
Sorry, I'm still able to reproduce the problems. My environment:
- Site Editor;
- Gutenberg `trunk`;
- Layout Grid `trunk`;
- WordPress 5.9.

This is my template:

```html
<!-- wp:group {"align":"full","layout":{"inherit":false,"contentSize":"70%"}} -->
<div class="wp-block-group alignfull"><!-- wp:paragraph -->
<p>Hey</p>
<!-- /wp:paragraph --></div>
<!-- /wp:group -->

<!-- wp:group {"align":"full","layout":{"inherit":false,"contentSize":"70%"}} -->
[...]

PR #258, opened 2022-01-11
Change layout grid to not be fullwide by default.

jasmussen:
Per a conversation in https://github.com/Automattic/block-experiments/issues/249#issuecomment-1008612929, it isn't always obvious why a layout grid block does not appear to be full-wide, if it's inside a group block that's not full-wide. 

I suspect that there are upstream improvements that can be made, but in the mean time, we can make full-wide be an opt-in. What do you think?

Screenshot 2022-01-11 at 12 18 35

### DIFF
--- a/blocks/layout-grid/src/index.js
+++ b/blocks/layout-grid/src/index.js
@@ -89,10 +89,6 @@ export function registerBlock() {
 			],
 		},
 		attributes: {
-			align: {
-				type: 'string',
-				default: 'full',
-			},
 			gutterSize: {
 				type: 'string',
 				default: 'large',


zaguiini:
It works as expected. My concern with it is that it will affect existing Layout Grids that do not have alignment settings set up😕 

Is there a way to affect only new instances of Layout Grid?

jasmussen:
> My concern with it is that it will affect existing Layout Grids that do not have alignment settings set up

Existing content shouldn't be affected by this change. Unless I'm mistaking something, this PR just changes the _preset_ for what you insert. Which means existing content, inserted with the full-wide default have that property saved into the content at this point, and should look unchanged. This default would affect only new content inserted. Even existing patterns should work as they do today.

zaguiini:
It's only broken inside the editor. The live site is fine. Reference:

https://user-images.githubusercontent.com/26530524/148995805-2df24d0b-a833-417c-9e61-976fd28ca5ee.mov

jasmussen:
Thanks for thorough testing. It's a bit hard from the video to tell which theme you're using, and how the full-wide behavior is meant to work. I.e. if you repeat the same steps in the video, but with a fullwide image and a non fullwide paragraph inside a group, instead of the layout grid, what happens?

jasmussen:
> To demonstrate even better. The first grid was inserted running the trunk version and the last one by using this branch's version:

Right, so the first grid is full wide upon insertion, the next grid is not full wide on insertion, you have to _toggle_ it fullwide using the alignments button. That was the desired change, right?

jasmussen:
Thanks for the patience!

It isn't clear to me how to best move forward. As far as I can tell, everything is technically working as it's meant to with regards to how `Layout` works in the block editor (i.e. that the group provides content, wide and full width metrics for all child blocks). The primary goal of this PR was to change the default behavior for _new_ layout grids so they default to content-width rather than full-width, and it should be technically impossible for this change to affect existing content since alignments are written into the markup and not dynamically sourced:
layout

It's very possible I'm missing a key bit here. But best I can tell, #249 details the potentially confusing fact that _a full-width layout grid inside a content-width block will only be as wide as the group_. This PR can only mitigate that in not defaulting the layout grid to full width for new content. But since this is expected behavior, if it isn't ideal, it might need to be brought up in [the core project](https://github.com/WordPress/gutenberg/). What do you think?

zaguiini:
Hi @jasmussen,

I'm fine with how the code looks and behaves. The only thing that I think might confuse the users is that the layout grid in the editor doesn't match the looks of the live site. As you can see in the screenshots of my last comment, I've shown that inside the editor, the layout grid block is being affected by the wrapping group. But that only happens inside the editor.

I'm not sure if it's possible to fix it. If it's not, then I'm fine with how this PR looks and we might create a follow-up to fix this problem. If it is, I'm happy to contribute!

zaguiini (approved):
Testing this again. On https://github.com/Automattic/block-experiments/pull/258#issuecomment-1010216992, what I did was:

- use `master` version of BE;
- add a layout grid and wrap it within a group block with layout constraints;
- save and open the front-end.

Both the editor and the front-end look the same.

Now, if I:
- use this PR's branch;
- reload the page.

In the editor, we're going to see the effects of the change in the existing layout grid. But that's only in the editor -- the live site looks fine. This does not look like an issue with this PR. It must be a higher-level problem with Gutenberg, so it's not related to the changes. @jasmussen said:

>Indeed. I do still think something higher level is going on here that warrants fixing in the block editor itself, i.e if you did those same steps, but with a full width columns block instead of the layout grid, would the problem be the same?
Nevertheless it seems okay to land this one as a bandaid.

And I agree. I'm approving this one, but I'd like to have one extra pair of eyes here. @Automattic/cylon, please!

jeyip:
@zaguiini 

I'm seeing the divergent behavior between the editor and the live site you described in [this comment](https://github.com/Automattic/block-experiments/pull/258#issuecomment-1011361738) without having this branch checked out.

To be clear, on my version of `trunk`, the layout grid already looks messed up in the editor and not the live site. I don't believe what you're seeing is an issue with Joen's branch.

jeyip (approved):

jasmussen:
Hey folks, thanks for the patience. In order to expedite future fixes, especially ones where folks like myself are lacking some of the context for the fix in question, we've put together release instructions for how to build and release a version of the plugin: https://github.com/Automattic/block-experiments#releasing-an-individual-block

Will you be able to merge this PR and make a release? Otherwise let us know and we can help!

PR #276, opened 2022-06-10
Layout Grid - Fix preview buttons in post editor

Addison-Stavlo:
Resolves  https://github.com/Automattic/wp-calypso/issues/62485

We can not rely on absence of the `core/edit-site` store for determining which device type functions should be called. Previously this did work, as the `core/edit-site` store was not available in the post editor while the `core/edit-post` store was available in both editors. In some cases, the `core/edit-site` store is present in the post editor context, causing errors with the device preview feature of this block when used in the post editor (as it tries to fire site editor functions instead of post editor ones). 

Here, we simply check for the edit `#edit-site-editor` container selector to determine which store to use for syncing the preview device type.

# Testing instructions
* Run this PR and build the block-experiments plugin, sync and enable it on a WordPress site.
* Add a Layout Grid block and initialize it with some column selection.
* In the block inspector sidebar, try the 3 device type buttons for Desktop, Tablet, and Mobile.
* Verify these set the preview device as expected in the editor canvas in both the Site Editor as well as the page/post Editor.

### DIFF
--- a/blocks/layout-grid/src/grid/higher-order.js
+++ b/blocks/layout-grid/src/grid/higher-order.js
@@ -127,7 +127,7 @@ export function withSetPreviewDeviceType() {
 				);
 
 				if ( isSiteEditor ) {
-					dispatch(
+					return dispatch(
 						'core/edit-site'
 					)?.__experimentalSetPreviewDeviceType( type );
 				}
--- a/blocks/layout-grid/src/grid/higher-order.js
+++ b/blocks/layout-grid/src/grid/higher-order.js
@@ -122,17 +122,19 @@ export function withSetPreviewDeviceType() {
 	return withDispatch( ( dispatch ) => {
 		return {
 			setPreviewDeviceType( type ) {
-				const {
-					__experimentalSetPreviewDeviceType: setSiteEditorDeviceType,
-				} = dispatch( 'core/edit-site' );
+				const isSiteEditor = document.querySelector(
+					'#edit-site-editor'
+				);
 
-				if ( setSiteEditorDeviceType ) {
-					setSiteEditorDeviceType( type );
-				} else {
+				if ( isSiteEditor ) {
 					dispatch(
-						'core/edit-post'
-					).__experimentalSetPreviewDeviceType( type );
+						'core/edit-site'
+					)?.__experimentalSetPreviewDeviceType( type );
 				}
+
+				dispatch(
+					'core/edit-post'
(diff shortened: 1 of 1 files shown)


blocks/layout-grid/src/grid/higher-order.js
```diff
@@ -122,10 +122,19 @@ export function withSetPreviewDeviceType() {
...
-				const { __experimentalSetPreviewDeviceType } =
-					dispatch( 'core/edit-site' ) ||
-					dispatch( 'core/edit-post' );
-				__experimentalSetPreviewDeviceType( type );
+				const isSiteEditor = document.querySelector(
+					'#edit-site-editor'
+				);
+
+				if ( isSiteEditor ) {
+					dispatch(
```
zaguiini, line 130: We're missing a short-circuit here (`return`) in case we are inside the post editor -- otherwise it's going to dispatch two actions 😸

Apart from that, the code looks good 👍 

I wonder if we could extract this `isSiteEditor` variable to a function or a constant elsewhere so we could have a single source of truth and more reliability. Or, I don't know if we have a function already that performs this kind of check...

Addison-Stavlo: > We're missing a short-circuit here (return) in case we are inside the site editor -- otherwise, it's going to dispatch two actions 😸

Oh, good catch! Updated that. It shouldn't have had any adverse side effects as using that post editor dispatch from the site editor should have no ill effect, but OMG once you pointed it out it smelled so stinky. 😆 

> I wonder if we could extract this isSiteEditor variable to a function or a constant elsewhere so we could have a single source of truth and more reliability. Or, I don't know if we have a function already that performs this kind of check...

I don't think there are any other functions like that in block experiments or need for them (but i could be wrong I don't work in this repo much), but it's not a bad idea.  I tried searching for something in the gutenberg stores that might supply this info but didn't find anything.

We also do a similar check for tracking in calyso's wpcom-block-editor here (https://github.com/Automattic/wp-calypso/blob/85e2d342cd9aa1aaf8fcfb8295928f71811d3be1/apps/wpcom-block-editor/src/wpcom/features/utils.js#L10), but id leave packaging that to be exported/imported and shared in both repos as a welcome follow up.

zaguiini: Definitely 👍

zaguiini (approved):

zaguiini (approved):

PR #298, opened 2023-02-10
Sketch: Update save to store svg in post content instead of rendering dynamically

ebinnion:
This is a first pass at updating the Sketch block to store the SVG markup in the post content as opposed to rendering the SVG dynamically on post generation. The reasoning for this is that we can provide a better fallback for when the Sketch block does not exist.

Why does this matter?

We are considering how to deploy the Sketch block to WordPress.com. One option here is to simply ship the plugin to WordPress.com. The downside to this is that we also have to consider Atomic sites. There are options for this. The simplest one seems to be to install the Sketch plugin when they'd like to update their block, but otherwise to default to rendering the block.

### DIFF
--- a/blocks/sketch/index.php
+++ b/blocks/sketch/index.php
@@ -92,4 +92,3 @@ function get_svg_path_from_stroke( $stroke ) {
 }
 
 add_action( 'init', __NAMESPACE__ . '\set_render_callback' );
-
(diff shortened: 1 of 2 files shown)


ebinnion:
Closing for now in favor of another approach.

PR #319, opened 2023-03-03
Sketch block: Add default and custom color palettes

oskosk:
<table>
<tr>
	<td>
image
	<td>image
</td>
</tr>
<tr>
	<td>Before
	<td>After
</table>

* Adds default and custom color palettes
* Fixes warning about deprecated isAlternate prop on Popover component 
* Adds padding to color palette popover
* Fixes selected value prop name passed to ColorPalette

### DIFF
--- a/blocks/sketch/editor.scss
+++ b/blocks/sketch/editor.scss
@@ -7,4 +7,12 @@
 			min-width: initial;
 		}
 	}
+	&__color-palette-popover {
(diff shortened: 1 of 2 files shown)


oskosk:
It seems like these changes don't work with WordPress 6.1.1. Only with Gutenberg 15.2.3. I'm checking how to address this

PR #329, opened 2023-06-21
Layout Grid: Don't write back to `props` in icon

### LINKED ISSUE #328: Layout grid is completely broken
The Layout Grid block is completely broken when using Gutenberg trunk, inserting it results in a "This block has encountered an error and cannot be previewed', and just opening the inserter crashes Gutenberg completely.

Both errors have the same message:

> Uncaught TypeError: Cannot add property width, object is not extensible at icon 

I guess this is because [the layout grid icon is writing back to the `props` parameter](https://github.com/Automattic/block-experiments/blob/4e117dad6e4eaaf88a0560cbc6d0be0bb48e8c05/blocks/layout-grid/src/icons.js#L9-L10C3). I'm unsure why this is only a problem with GB trunk though, I can't reproduce on GB16.

dsas:
Writing to the `props` passed causes an error if it doesn't already have width and height properties as the object isn't extensible.

This is fixed by copying to an intermediary object and then using that.

This error is fatal - crashes the whole of GB when using current GB trunk (15d644f2)

### Testing steps
 1. Checkout this branch locally
 2. `nvm use && yarn install`
 3. Build these changes: `yarn plugin layout-grid && yarn bundle`
 4. Get the built plugin in your test environment
    1. For wpcom unzip the plugin: `cd bundles && unzip jetpack-layout-grid.zip && cd jetpack-layout-grid`
    2. Then copy it up`scp -r * wpcom-sandbox:public_html/wp-content/plugins/gutenberg-blocks/jetpack-layout-grid/`
 5. Make your test environment use the latest version of GB, for wpcom see the FGs gutenberg-plugin-on-wpcom

Closes #328

### DIFF
--- a/blocks/layout-grid/src/icons.js
+++ b/blocks/layout-grid/src/icons.js
@@ -5,16 +5,17 @@
 import { Path, SVG } from '@wordpress/components';
 
 export const GridIcon = ( props ) => {
+	const iconProps = {...props};
 	if ( props.size ) {
-		props.width = props.size;
-		props.height = props.size;
+		iconProps.width = props.size;
+		iconProps.height = props.size;
 	}
 
 	return (
 		<SVG xmlns="http://www.w3.org/2000/svg"
 			width="24" height="24"
 			viewBox="0 0 24 24"
-			{ ...props }
+			{ ...iconProps }
 		>
 			<Path d="M19 6H6c-1.1 0-2 .9-2 2v9c0 1.1.9 2 2 2h13c1.1 0 2-.9 2-2V8c0-1.1-.9-2-2-2zm-7.5 11.5H6c-.3 0-.5-.2-.5-.5V8c0-.3.2-.5.5-.5h5.5v10zm4 0H13v-10h2.5v10zm4-.5c0 .3-.2.5-.5.5h-2v-10h2c.3 0 .5.2.5.5v9z" />
 		</SVG>
--- a/blocks/layout-grid/src/icons.js
+++ b/blocks/layout-grid/src/icons.js
@@ -5,7 +5,7 @@
 import { Path, SVG } from '@wordpress/components';
 
 export const GridIcon = ( props ) => {
-	const iconProps = {...props};
+	const iconProps = { ...props };
 	if ( props.size ) {
 		iconProps.width = props.size;
 		iconProps.height = props.size;


dsas:
Note that the tests fail, but then they failed before and it looks unrelated so...

blocks/layout-grid/src/icons.js
```diff
@@ -5,16 +5,17 @@
 import { Path, SVG } from '@wordpress/components';
 
 export const GridIcon = ( props ) => {
+	const iconProps = {...props};
 	if ( props.size ) {
-		props.width = props.size;
-		props.height = props.size;
+		iconProps.width = props.size;
+		iconProps.height = props.size;
 	}
```
rcrdortiz, line 12: When the size isn't set is it equivalent to having the width and height set to 0? If so we could simplify the code by leveraging the optional chaining operator and null coalescing operator and end up with (if zero isn't good we can think of another default value):
```
const iconProps = {
    width: props?.size ?? 0,
    height: props?.size ?? 0,
    ...props
};
```

dsas: Omitting width and height would mean they get set to 'auto' (resolving `auto` is a whole can of worms but we don't need to worry about that). 

I don't think your suggestion is really significantly simpler. I'll change it if you disagree though.

rcrdortiz: Given your explanation that omitting width and height would set them to 'auto', it indeed seems your implementation has a different behavior than the one I suggested. I was under the impression that the absence of these properties would default to '0', not 'auto'. Therefore, if 'auto' is the intended behavior when props.size is not provided, your current implementation seems appropriate.

As a general rule, small steps towards simpler code can aggregate into a significant simplification of our code base over time, that's why I usually insist on applying small or marginal simplifications.

blocks/layout-grid/src/icons.js
```diff
@@ -5,16 +5,17 @@
 import { Path, SVG } from '@wordpress/components';
 
 export const GridIcon = ( props ) => {
+	const iconProps = {...props};
```
mmtr, line 8: Minor code style change
```suggestion
	const iconProps = { ...props };
```

mmtr (approved):

PR #332, opened 2023-07-24
Motion Background: Fix function redeclaration error

ajlende:
## Description

In WordPress 6.3, the `enqueue_block_assets` was getting called twice causing the function inside to be redeclared.

This PR fixes the issue by moving the script enqueues into the `render_callback`. The code is also refactored to be a bit more like how [Jetpack blocks](https://github.com/Automattic/jetpack/blob/trunk/projects/plugins/jetpack/extensions/blocks/calendly/calendly.php) are structured.

## Testing Instructions

1. Open the site editor and post editor.
2. See that there are no errors.

### DIFF
--- a/blocks/motion-background/index.php
+++ b/blocks/motion-background/index.php
@@ -1,39 +1,52 @@
 <?php
 
-add_action( 'init', function() {
-	register_block_type( 'a8c/motion-background', [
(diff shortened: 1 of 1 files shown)


ajlende:
Closing in favor of #333.

PR #337, opened 2023-11-06
Event Block: Update deprecated APIs

ajlende:
Some of the experimental APIs used when this block was created were stabilized or removed in later versions of Gutenberg, so they have been replaced or removed to work with WordPress 6.4.

Also added a note about this block no longer being actively developed.


scruffian:
Thanks for taking care of this

PR #340, opened 2023-12-21
Update node to version 20

fluiddot:
This PR is a follow-up on the Gutenberg PR https://github.com/WordPress/gutenberg/pull/56331 to update Node to the same version used in Gutenberg.

More information can be found in:
https://make.wordpress.org/core/2023/12/20/updating-wordpress-to-use-more-modern-versions-of-node-js-npm-2/


package.json
```diff
@@ -30,7 +30,7 @@
 		"@wordpress/icons": "^6.1.1",
 		"@wordpress/jest-preset-default": "^7.1.3",
 		"@wordpress/prettier-config": "^1.1.1",
-		"@wordpress/scripts": "^19.2.3",
+		"@wordpress/scripts": "^26.19.0",
```
fluiddot, line 33: It's unclear the potential side-effects of updating this package. The fact that this repository doesn't have CI checks gives low confidence in introducing changes. In any case, I presume that this package is mainly used for the development workflow, so probably won't introduce regressions.

ajlende: Each plugin is manually released separately, so there at least wouldn't be any immediate problems if one of them no longer builds properly.

I ran `yarn everything` to build all the plugins, and didn't see any issues that didn't exist before this change.

ajlende (approved):
The blocks that I maintain here look fine: Bauhaus Centenary, Event, Model Viewer, Starscape, and Waves.

It looks like the waves and book blocks are a bit broken, but it's probably because they're using old `__experimental` APIs in Gutenberg which have been deprecated and removed, not because of the changes in this PR.

I also had to make sure that I ran a manual `yarn clean` before `yarn start` for development because there were some extra files floating around causing errors.

It's probably worthwhile having someone more familiar with the layout grid test that one out since it's more widely used. But I'll give it one check of approval for the things that I maintain.

package.json
fluiddot: Great, thank you very much @ajlende for the information and for checking that all plugins can be built 🙇 .

fluiddot:
> The blocks that I maintain here look fine: Bauhaus Centenary, Event, Model Viewer, Starscape, and Waves.
> It looks like the waves and book blocks are a bit broken, but it's probably because they're using old `__experimental` APIs in Gutenberg which have been deprecated and removed, not because of the changes in this PR.

Thanks @ajlende for testing some of the blocks. Following https://github.com/Automattic/block-experiments/pull/340#discussion_r1434527639, since each plugin is released individually, in potential next releases we'd need to test if any regressions have been introduced.

> I also had to make sure that I ran a manual `yarn clean` before `yarn start` for development because there were some extra files floating around causing errors.

Ah, interesting. I didn't experience this but I presume that changing the node version or dependency updates might be related.

> It's probably worthwhile having someone more familiar with the layout grid test that one out since it's more widely used. But I'll give it one check of approval for the things that I maintain.

Ok, thanks 🙇 ! I'll share a list with the remaining blocks to check and reference the last contributors who worked on them.

fluiddot (commented):
As shared in https://github.com/Automattic/block-experiments/pull/340#discussion_r1434527639, seems that all blocks can be built. However, I noticed that all unit tests are failing. Some of them don't pass either in `master` but before merging this PR, we should at least fix the unit tests that were succeeding before this change.

The error I encountered is the following:
```
The error below may be caused by using the wrong test environment, see https://jestjs.io/docs/configuration#testenvironment-string.
    Consider using the "jsdom" test environment.

    ReferenceError: window is not defined

      at Object.<anonymous> (node_modules/@wordpress/jest-preset-default/scripts/setup-globals.js:2:14)
```

fluiddot:
Following https://github.com/Automattic/block-experiments/pull/340#pullrequestreview-1794216776, I tried to address the issue when running tests but seems more complex than I originally thought. In the newer version of `@wordpress/scripts`, the Jest version is updated and leads to several errors that might imply updating the Jest configuration and possibly some of the WordPress packages. I tried to force the resolution of `eslint-plugin-jsdoc` (the package that fails when installing dependencies with node 20) to a newer version, but still Jest fails with the following error:
```
Jest encountered an unexpected token

    This usually means that you are trying to import a file which Jest cannot parse, e.g. it's not plain JavaScript.

    By default, if Jest sees a Babel config, it will use that to transform your files, ignoring "node_modules".

    Here's what you can do:
     • If you are trying to use ECMAScript Modules, see https://jestjs.io/docs/en/ecmascript-modules for how to enable it.
     • To have some of your "node_modules" files transformed, you can specify a custom "transformIgnorePatterns" in your config.
     • If you need a custom transformation specify a "transform" option in your config.
     • If you simply want to mock your non-JS modules (e.g. binary assets) you can stub them out with the "moduleNameMapper" config option.

    You'll find more details and examples of these config options in the docs:
    https://jestjs.io/docs/en/configuration.html

    Details:

    <root-path>/block-experiments/node_modules/@wordpress/i18n/node_modules/memize/dist/index.js:160
    export { memize as default };
    ^^^^^^

    SyntaxError: Unexpected token 'export'

      at Runtime.createScriptFromCode (node_modules/jest-runtime/build/index.js:1350:14)
      at Object.<anonymous> (node_modules/@wordpress/i18n/build/@wordpress/i18n/src/sprintf.js:4:1)
      at Object.<anonymous> (node_modules/@wordpress/i18n/build/@wordpress/i18n/src/index.js:1:1)
      at Object.<anonymous> (node_modules/@wordpress/keycodes/build/@wordpress/keycodes/src/index.js:15:1)
      at Object.<anonymous> (node_modules/@wordpress/compose/build/hooks/use-constrained-tabbing/@wordpress/compose/src/hooks/use-constrained-tabbing/index.js:4:1)
      at Object.<anonymous> (node_modules/@wordpress/compose/build/@wordpress/compose/src/index.js:21:1)
      at Object.<anonymous> (node_modules/@wordpress/data/build/redux-store/@wordpress/data/src/redux-store/index.js:13:1)
      at Object.<anonymous> (node_modules/@wordpress/data/build/@wordpress/data/src/registry.js:14:1)
      at Object.<anonymous> (node_modules/@wordpress/data/build/@wordpress/data/src/default-registry.js:4:1)
      at Object.<anonymous> (node_modules/@wordpress/data/build/@wordpress/data/src/index.js:9:1)
```

Until this gets resolved, I think we should pause this PR so I'm going to revert it to draft. As a follow-up, I'll open an issue with this information.

PR #344, opened 2025-02-26
Update nvm to match Gutenberg version and upgrade yarn packages

jeryj:
I updated the `nvm` node version to 20 to match what [is used by Gutenberg](https://github.com/WordPress/Gutenberg/blob/trunk/.nvmrc) and then ran `yarn upgrade --latest`

### Testing instructions
- `nvm use`
- Test the watcher: `yarn start`
- Make a simple change to the layout grid and make sure it builds. Leave the change there for the moment. You'll want to confirm its presence in the final build.
- Run tests: `yarn jest`: You will get warning around core blocks already being registered. This is not ideal but OK. The tests will fail. I can't tell if they were failing before as well (since the build to test it doesn't work)
- Make a new build: 
   - `yarn plugin layout-grid`
   - `yarn bundle`
   - Check the bundle at `bundles/layout-grid.zip` to make sure your change is included.


jest.config.js
```diff
@@ -1,28 +1,8 @@
...
-		'./config/jest-before.js',
-	],
-	testMatch: [
-		'**/__tests__/**/*.[jt]s',
-		'**/test/*.[jt]s',
-		'**/?(*.)test.[jt]s',
-	],
-	testPathIgnorePatterns: [ '/node_modules/', '<rootDir>/wordpress/' ],
-	timers: 'fake',
+	preset: '@wordpress/jest-preset-default',
```
jeryj, line 2: Most of the old file was already set-up in @wordpress/jest-preset-default.

scruffian:
I tried it using node 16.17.0. I had to run `python3 -m pip install setuptools` to get `distutils` installed, and then the build seems to work.

scruffian:
I'm not sure how to run the tests - I tried `yarn run jest` which seemed to work but I got a lot of failures. 

When I tried using the block it seemed to work as expected.

scruffian:
Transforming the block into a columns block doesn't work well.

jeryj:
> Transforming the block into a columns block doesn't work well.

Is this a new bug or is it working the same as before?

scruffian (approved):
I followed the testing instructions and the build seems to work. Given that we want to retire this block I don't think we should fix the tests.

Issue #327, opened 2023-05-26
Layout Grid behaving weirdly with latest Gutenberg


tellthemachines:
The Layout Grid resize handles are acting up when latest version of Gutenberg is enabled:

https://github.com/Automattic/block-experiments/assets/8096000/331e75e9-a97e-4b5b-842a-09704b6db449

This is most likely due to a change on the Gutenberg side, but it's hard to debug as I'm not really familiar with this block.

Also, trying to add Layout Grid to a post in my local dev environment with Gutenberg trunk results in the whole editor crashing with `Uncaught TypeError: Cannot add property width, object is not extensible`, making it even harder to get to the bottom of the issue 😅

philbraun:
I get the same strange behaviour, and Layout Grids not being saved, but only when trying to use it inside the Site Editor in a Block Theme. Maybe the Site Editors Preview size setting conflicts with LG's Responsive Breakpoints settings?

PR #345, opened 2025-02-26
Fix grid resize listeners

jeryj:
Fixes https://github.com/Automattic/block-experiments/issues/327

The ResizeGrid component was not emitting the onMouseUp event. Attaching it to the div allows the mouseUp to be emitted if you're within the block. This is not a full fix though. The mouseUp event should be emitted anywhere you release it, not just within the block.

Maybe it would be worth updating this to the [Resizeable box component](https://wordpress.github.io/gutenberg/?path=/docs/components-resizablebox--docs) and only enable the left/right handles. 

**Before**

https://github.com/user-attachments/assets/021bb02a-6291-4e0d-9429-aaad1d00ce98

**After**

https://github.com/user-attachments/assets/d55707c7-38bc-4dac-9e0b-c5a227243f25

### DIFF
--- a/blocks/layout-grid/src/grid/resize-grid/index.js
+++ b/blocks/layout-grid/src/grid/resize-grid/index.js
@@ -41,7 +41,18 @@ class ResizeGrid extends Component {
 		const { totalColumns, layoutGrid } = this.props;
 		const start = layoutGrid.getStart( this.state.resizingColumn );
 		const span = layoutGrid.getSpan( this.state.resizingColumn );
-		const nearest = Math.min( totalColumns, Math.max( 0, findNearest( this.containerRef.current, this.getMouseX( mouse ), direction, totalColumns ) ) );
+		const nearest = Math.min(
+			totalColumns,
+			Math.max(
+				0,
+				findNearest(
+					this.containerRef.current,
+					this.getMouseX( mouse ),
+					direction,
+					totalColumns
+				)
+			)
+		);
 
 		if ( direction === 'left' ) {
 			if ( nearest === start ) {
@@ -85,7 +96,11 @@ class ResizeGrid extends Component {
 		const { width } = this.state;
 		const handleWidth = optionalWidth > 0 ? optionalWidth : width;
 
-		return offset - this.containerRef.current.getBoundingClientRect().left - ( ( handleWidth ) / 2 );
+		return (
+			offset -
+			this.containerRef.current.getBoundingClientRect().left -
+			handleWidth / 2
+		);
 	}
 
 	getAdjustedTop( offset ) {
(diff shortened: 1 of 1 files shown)


blocks/layout-grid/src/grid/resize-grid/index.js
```diff
@@ -41,7 +41,18 @@ class ResizeGrid extends Component {
...
+			Math.max(
+				0,
+				findNearest(
+					this.containerRef.current,
+					this.getMouseX( mouse ),
+					direction,
+					totalColumns
+				)
+			)
+		);
```
scruffian, line 55: This looks like its just a formatting change.

blocks/layout-grid/src/grid/resize-grid/index.js
```diff
@@ -170,28 +192,46 @@ class ResizeGrid extends Component {
...
-	onMouseUp = ev => {
-		this.setState( { resizingColumn: -1 } );
+	onMouseUp = ( ev ) => {
+		if ( this.state.resizingColumn === -1 ) {
+			return;
+		}
 
+		this.setState( { resizingColumn: -1 } );
 		document.removeEventListener( 'mousemove', this.onMouseMove );
-		document.removeEventListener( 'mouseup', this.onMouseUp );
```
scruffian, line 179: This and line 181 looks like the only change that isn't formatting.

scruffian:
I'm not sure how to test this - I struggled to get the block to load in my environment.

blocks/layout-grid/src/grid/resize-grid/index.js
jeryj: Yeah, that's correct. The first commit on this branch is just linting. It's going to be a thing for the whole package, as the npm versions were really out of date.

jeryj:
The [development environment setup](https://github.com/Automattic/block-experiments/?tab=readme-ov-file#development) and yarn instructions worked for me

jeryj:
The block actually works fine. The error is in Gutenberg's iframe bubbling. It doesn't allow the `mouseup` event to bubble to the document. [Adding `mouseup` to this array of events fixes the error.](https://github.com/WordPress/gutenberg/blob/update/navigation-custom-appender/packages/block-editor/src/components/iframe/index.js#L79)

mikachan (approved):
Nice, this works well! I can see the same as you describe on both `master` and this PR - and the resize handles work better on this PR:


https://github.com/user-attachments/assets/498e23ad-ab3d-40ac-bcea-9ecf2bcc677b


> Maybe it would be worth updating this to the [Resizeable box component](https://wordpress.github.io/gutenberg/?path=/docs/components-resizablebox--docs) and only enable the left/right handles.

This sounds like a good follow-up, but I think this current fix is fine to bring in so we can fix https://github.com/Automattic/block-experiments/issues/327 quickly.

PR #346, opened 2025-03-25
Fix layout grid listeners

jeryj:
Redo of the fix for https://github.com/Automattic/block-experiments/pull/345. That one got merged too early. This is built off of [`update/nvm`](https://github.com/Automattic/block-experiments/pull/344), but shouldn't get merged until `update/nvm` is merged.

This PR includes linting changes for the file. The actual fix is just using `target.ownerDocument` instead of `document` because `document` won't work when the gutenberg editor is iframed.

### DIFF
--- a/blocks/layout-grid/src/grid/resize-grid/index.js
+++ b/blocks/layout-grid/src/grid/resize-grid/index.js
@@ -41,7 +41,18 @@ class ResizeGrid extends Component {
 		const { totalColumns, layoutGrid } = this.props;
 		const start = layoutGrid.getStart( this.state.resizingColumn );
 		const span = layoutGrid.getSpan( this.state.resizingColumn );
-		const nearest = Math.min( totalColumns, Math.max( 0, findNearest( this.containerRef.current, this.getMouseX( mouse ), direction, totalColumns ) ) );
+		const nearest = Math.min(
+			totalColumns,
+			Math.max(
+				0,
+				findNearest(
+					this.containerRef.current,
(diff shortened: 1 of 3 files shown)


blocks/layout-grid/src/grid/resize-grid/index.js
```diff
@@ -114,15 +129,24 @@ class ResizeGrid extends Component {
...
+			( ev.button === 0 || ev.touches ) &&
+			( target.dataset.resizeRight || target.dataset.resizeLeft )
+		) {
 			this.block = target.closest( '.wp-block' );
 
-			const { height, right, left, top } = this.block.getBoundingClientRect();
+			// Get the document that contains the target element. We need to use target.ownerDocument
+			// because the Gutenberg editor is iframed, so some events are prevented from bubbling to the document.
+			// https://github.com/WordPress/gutenberg/blob/trunk/packages/block-editor/src/components/iframe/index.js#L79
+			this.targetDocument = target.ownerDocument;
```
jeryj, line 146: Change.

jeryj (commented):
These lines are the ones that have actual changes (not linting)

scruffian (approved):
This fixes the problem with handles not being released. It's difficult to be certain that nothing else broke because the tests don't work. I have tested the block manually and I can't find any regressions. I suggest that we deploy this change to WordPress.com and watch to see if anything comes up. If it does we'll have to revert.

Approving on the assumption that you also update the version number and README files.

